PR #15, opened 2018-05-10
Fixes small issue in front end with clicking land button

lukebatchelor:
Small fix to misnamed qs variable, and adding a function to window.

Rest is just prettier im pretty sure

### DIFF
--- a/src/hosts/BitbucketAdapter.js
+++ b/src/hosts/BitbucketAdapter.js
@@ -40,10 +40,15 @@ const BitbucketAdapter = (config: Config) => {
       parentCommentId: ?string,
       message: string
     ) {
-      let data = { content: message };
+      let data = {
+        content: message
+      };
 
       if (parentCommentId) {
-        data = { ...data, parent_id: String(parentCommentId) };
+        data = {
+          ...data,
+          parent_id: String(parentCommentId)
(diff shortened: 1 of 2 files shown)


src/static/bitbucket/index.js
```diff
@@ -82,10 +82,10 @@ function getCurrentState() {
 // Fetches the user, repo and id vars
 function getQueryStringVars() {
   const qs = window.location.search.substring(1);
-  return queryString.parse(queryString);
+  return queryString.parse(qs);
 }
 
-function wantToMergeClicked() {
+function wantToMergeClicked1() {
```
lukebatchelor: Gross hack, ignore it!

src/hosts/BitbucketAdapter.js
```diff
@@ -88,14 +103,25 @@ const BitbucketAdapter = (config: Config) => {
         } merged  by Landkid after a successful build rebased on Master`,
         merge_strategy: 'merge_commit'
       };
-      Logger.info({ pullRequestId, endpoint }, 'Merging pull request');
```
lukebatchelor: Not sure why prettier has decided to be stricter on these? We've always had prettier running...

PR #17, opened 2018-05-11
Adds landWhenAble functionality

lukebatchelor:
All functionality there.

Just using a simple array to track them atm.

Definitely want to revisit the State in Runner at some point.

Added a bunch of TODO's around the place.

Only known but that I'll get to soon is that refreshing the page after clicking mergeWhenAble will show you that you are not in the queue.

In terms of testing, I'll need to do a bunch around making sure that duplicated webhooks and combinations of landRequests + mergeWhenAble requests don't lead to any weirdness

### DIFF
--- a/src/static/bitbucket/index.js
+++ b/src/static/bitbucket/index.js
@@ -3,13 +3,24 @@ const queryString = require('qs');
 
 const endpoint = window.location.origin;
 
-const landButtonView = () => {
+function createBtn(btnText, appearance, onClickFn) {
+  return `<button type="button" class="ak-button ak-button__appearance-${
+    appearance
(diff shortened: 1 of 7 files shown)


src/static/bitbucket/index.js
```diff
@@ -73,9 +90,17 @@ const errorCreatingLandRequestView = err => {
   </div>`;
 };
 
-function getCurrentState() {
+function getCurrentStateAndSettings() {
   return fetch(`${endpoint}/api/current-state`)
     .then(resp => resp.json())
+    .then(state =>
+      fetch(`${endpoint}/api/settings`)
```
lukebatchelor: Would be great to just async await this stuff...

PR #20, opened 2018-05-25
Adds retrying when trying to merge

lukebatchelor:
I also finally fixed my prettier problems, so I'll slowly be fixing files as I touch them.

### DIFF
--- a/src/hosts/BitbucketAdapter.js
+++ b/src/hosts/BitbucketAdapter.js
@@ -1,27 +1,29 @@
 // @flow
+import axios from 'axios';
+import pRetry from 'p-retry';
 import type {
   HostAdapter,
   JSONValue,
   PullRequest,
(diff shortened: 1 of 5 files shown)


src/hosts/BitbucketAdapter.js
```diff
@@ -35,61 +37,55 @@ const BitbucketAdapter = (config: HostConfig) => {
...
         `${oldApiBaseUrl}/pullrequests/${pullRequestId}/comments/`,
         JSON.stringify(data),
-        axiosPostConfig
+        axiosPostConfig,
       );
 
       return response.data.comment_id;
     },
 
     async mergePullRequest(pullRequestId: string) {
```
lukebatchelor: This doesn't actually need to be async anymore tbh

PR #31, opened 2019-11-04
Admin build controls

gardnerjack:
Added buttons for **next** build, **cancel current** build, and **remove** from queue.  
Improved pause checkbox (previously it paused builds even after denying the confirm).  
Using an admin control causes the page to reload so the page is updated with the newest build information.

Added pause reason to the Bitbucket PR display.

Tabs file was becoming unwieldy so I split it up into a file for each Tab (under `components/tabs`). Also use `sessionStorage` to save the current tab since I was getting annoyed when refreshing the page reset it to the queue view. 

image
image
image


src/lib/Runner.ts
```diff
@@ -174,12 +174,12 @@ export class Runner {
     return state.get();
   };
 
-  public isPaused = async () => {
+  public isPaused = async (): Promise<string | false> => {
```
MarshallOfSound: `string | false` makes it hard to do a "falsey" check which can lead to bad assumption.  (Empty strings are falsey).

If you want to get the reason out of this method the return type should be.

```typescript
public getPauseState = async (): Promise<{
  isPaused: true;
  reason: string;
} | {
  isPaused: false;
}> => {}
```

src/static/current-state/components/QueueItem.tsx
```diff
@@ -129,59 +129,82 @@ const buildUrlFromId = (base: string, id: number) => `${base}/addon/pipelines/ho
...
-            </Lozenge>
-          </StatusItem>
-        ) : null}
-      </div>
-    </a>
-  );
+export class QueueItem extends React.Component<QueueItemProps> {
+  handleRemoveClick = () => {
+    fetch(`/api/remove/${this.props.request.id}`, { method: 'POST' })
+      .then(r => r.json())
```
MarshallOfSound: This should handle the failure case (`r.status !== 200`)

src/static/current-state/components/tabs/SystemTab.tsx
```diff
@@ -0,0 +1,146 @@
...
+      fetch('/api/unpause', { method: 'POST' }).then(() => {
+        this.setState({ paused: false });
+        location.reload();
+      });
+    }
+  };
+
+  onNextClick = () => {
+    fetch('/api/next', { method: 'POST' })
+      .then(r => r.json())
```
MarshallOfSound: Same here, this new API has a failure state.  The UI shouldn't be inconsistent with the returned status code

src/static/current-state/components/tabs/SystemTab.tsx
```diff
@@ -0,0 +1,146 @@
...
+      .then(() => location.reload());
+  };
+
+  onCancelClick = () => {
+    fetch('/api/cancel-current', { method: 'POST' })
+      .then(r => r.json())
```
MarshallOfSound: And here

lukebatchelor:
Sorry, just getting around to this now

PR #33, opened 2019-11-19
Banner message, display target branch, and improved Allowed Users layout

gardnerjack:
Database migration scripts for the `targetBranch` column in the `PullRequest` table, and creating a new table called `MessageStateTransition`.

New layout/styling for the Allowed Users section of the system tab, no longer breaks if the name is too long + Atlaskit select. (albeit changed slightly).

Display the target branch of a pull request on land request items that are waiting to queue, queued, or running. If the target branch is `master` the lozenge is blue, and if the name of the branch is long it hides the overflow and displays the full name on hover.

target-branch

New admin feature: displaying custom messages on pull requests, can be styled as either `default`, `warning`, or `error` messages.

default
warning
error


migrations/2_MessageStateTransitionTable.js
```diff
@@ -0,0 +1,34 @@
+module.exports = {
+  up: function(query, Sequelize) {
+    return query.createTable('MessageStateTransition', {
+      id: {
+        primaryKey: true,
+        type: Sequelize.UUID,
+        defaultValue: Sequelize.UUIDV4,
```
RajaBellebon: What is the difference between using UUID and UUIDV4?

migrations/2_MessageStateTransitionTable.js
```diff
@@ -0,0 +1,34 @@
...
+        allowNull: false,
+        type: Sequelize.STRING,
+      },
+      messageExists: {
+        allowNull: false,
+        type: Sequelize.BOOLEAN,
+      },
+      message: {
+        allowNull: true,
+        type: Sequelize.STRING({ length: 2000 }),
```
RajaBellebon: Is this a way to limit the length of the string?

src/lib/Runner.ts
```diff
@@ -190,6 +191,60 @@ export class Runner {
...
+    await MessageStateTransition.create<MessageStateTransition>({
+      senderAaid: user.aaid,
+      messageExists: true,
+      message,
+      messageType,
+    });
+  }
+
+  async removeBannerMessage(user: ISessionUser) {
+    await MessageStateTransition.create<MessageStateTransition>({
```
RajaBellebon: just to be sure - it will remove or hide the banner message?

src/static/bitbucket/components/App.tsx
```diff
@@ -1,7 +1,13 @@
 import * as React from 'react';
 import { proxyRequest } from '../utils/RequestProxy';
 
-export type AppState = {
+type BannerMessage = {
+  messageExists: boolean;
+  message: string | null;
```
RajaBellebon: if message is null it will be falsy so ```if (!message) {
        res.status(400).json({ error: 'Cannot send an empty message' });
      } ```will return 400 - do we want that ? Maybe we just do message: string

src/static/bitbucket/components/App.tsx
```diff
@@ -12,14 +18,32 @@ export type AppState = {
...
+  messageColours = {
+    default: 'transparent',
+    warning: 'orange',
+    error: 'red',
+  };
+
+  messageEmoji = {
+    default: '📢',
+    warning: '⚠️',
+    error: '❌',
```
RajaBellebon: There is an addition space after the X

RajaBellebon: maybe it is expected

src/static/bitbucket/components/App.tsx
RajaBellebon: I see below that you pass a null value for message

src/static/current-state/components/Lozenge.tsx
```diff
@@ -5,16 +5,18 @@ import { LozengeAppearance } from './types';
 
 let styles = css({
   boxSizing: 'border-box',
-  display: 'inline-flex',
+  display: 'block',
```
RajaBellebon: have you checked the rendering for`inline-block`?

src/static/current-state/components/QueueItem.tsx
```diff
@@ -150,22 +153,24 @@ export class QueueItem extends React.Component<QueueItemProps> {
...
 
     return (
       <a
         className={`${queueItemStyles} queue-item`}
-        href={request.buildId ? buildUrlFromId(bitbucketBaseUrl, request.buildId) : '#'}
+        href={buildId ? buildUrlFromId(bitbucketBaseUrl, buildId) : '#'}
       >
         <ak-grid layout="fluid">
-          <ak-grid-column size="10">
+          <ak-grid-column size={status.state === 'queued' ? 11 : 12}>
```
RajaBellebon: What was your reasoning for this change?

src/static/current-state/components/tabs/Messenger.tsx
```diff
@@ -0,0 +1,119 @@
...
+  messageColours = {
+    default: 'transparent',
+    warning: 'orange',
+    error: 'red',
+  };
+
+  messageEmoji = {
+    default: '📢',
+    warning: '⚠️',
+    error: '❌',
```
RajaBellebon: same as before

src/static/current-state/components/tabs/SystemTab.tsx
```diff
@@ -1,17 +1,11 @@
 import * as React from 'react';
 import { TabContent } from './TabContent';
-import { User } from '../User';
-import { PermissionControl } from '../PermissionControl';
-
-// sort by permssion descending (admin -> land -> read)
-function sortUsersByPermission(user1: IPermission, user2: IPermission) {
```
RajaBellebon: Where does it go?

src/static/bitbucket/components/App.tsx
gardnerjack: There actually isn't, it is misleading though, just how the emoji renders.

migrations/2_MessageStateTransitionTable.js
gardnerjack: UUID is the type of the column, and UUIDV4 is generator of UUIDs (which is why it's the default value, since it's used to create the UUID for each message). The same process is used with the other columns that use UUIDs.

src/lib/Runner.ts
gardnerjack: This will remove the banner message, as the structure is a transition between the a banner message existing and not existing. There would be no way of re-showing the message without creating a new one.

src/static/current-state/components/tabs/SystemTab.tsx
gardnerjack: This is moved into `src/static/current-state/components/tabs/AllowedUsers.tsx`

src/static/current-state/components/QueueItem.tsx
gardnerjack: The columns allow a button to be positioned in the right column of the queue item. However, the remove button is only available on items that are 'queued', so I've allowed items with other statuses that won't have the button to use the full width for their content.

src/static/current-state/components/Lozenge.tsx
gardnerjack: All three seem to render the same way.

RajaBellebon: ok cool :)

src/db/migrations/02__MessageStateTransitionTable.ts
```diff
@@ -0,0 +1,36 @@
+import { QueryInterface, DataTypes } from 'sequelize';
```
lukebatchelor: Hmmm.. Do you think this could have actually been added to the db.ts file directly since it's just a new table?
Happy to keep it here for now to prevent churn, but next time I might test that out. Better than having some db stuff be declared here and some there.

src/lib/Queue.ts
```diff
@@ -75,6 +76,7 @@ export class LandRequestQueue {
   public maybeGetStatusForQueuedRequestById = async (
     requestId: number,
   ): Promise<LandRequestStatus | null> => {
+    Logger.info('finding landrequeststatus', { requestId });
```
lukebatchelor: These two logs required? Can probably remove them I think

src/lib/Runner.ts
```diff
@@ -68,10 +69,10 @@ export class Runner {
 
         Logger.info('Land build now running', { running: landRequest.get() });
       } else {
-        Logger.info(
-          { ...isAllowedToLand, ...landRequest.get() },
-          'Land request is not allowed to land',
-        );
+        Logger.info('Land request is not allowed to land', {
+          ...isAllowedToLand,
```
lukebatchelor: Awesome catch on this one.

tools/hax.js
```diff
@@ -10,9 +10,27 @@ const typesPath = path.resolve(
...
+const sequelizeTypesPath = path.resolve(
+  __dirname,
+  '..',
+  'node_modules',
+  '@types',
+  'sequelize',
+  'index.d.ts',
+);
+
+if (fs.existsSync(passportTypesPath)) {
```
lukebatchelor: Ugh, I knew this would come back to haunt us.

lukebatchelor:
Awesome stuff @gardnerjack !

If I'd been more on top of this, I probably would have given more feedback earlier, so that's on me. I don't want to hold this back any more though, but some things I would have done differently:

* We could have probably made the banner message table be defined in the normal database setup right? I'm pretty sure it didnt need to be a migration. Just would have been nice to have as much stuff as possible declared in the same place. I'm not _super_ worried, though, just something for us to look out for next time.
* We probably didnt need that messageExists column in the table either right? I'd probably not worked around it using either an empty message string, or returning null when it doesnt exist (making the default values on those two getters a bit simpler too `getMessageState` and `getBannerMessage` could just return their simple shapes or null if needed.
* I'd probably keep the naming of this table / columns more consistent. Banner, Message, Messenger, etc. I probably would have stuck to just BannerMessage everywhere for consistency. Not a huge issue, and something we can probably clean up later, so Im happy to not have that block this.

Also going to merge with those two extra logs in there too, but they can be removed later.

Other than that though, awesome job on this. This is a huge step forwards and creates some new patterns we can build on for the future.

PR #34, opened 2019-11-22
Custom Pull Request checks/rules

gardnerjack:
Important files are:

- config.example.js
- src/bitbucket/BitbucketClient.ts
- src/lib/Runner.ts
- src/routes/bitbucket/proxy/index.ts
- src/types.ts

Everything else is just me changing some components into arrow functions for consistency's sake.

### DIFF
--- a/config.example.js
+++ b/config.example.js
@@ -35,10 +35,11 @@ module.exports = {
      *    approvals: string[] -- usernames of all real approvals
      *    permissionLevel: "read" | "land" | "admin" -- permission level of the user requesting /can-land
      *  }
-    rules: [
+     * Return true if the rule is passed and is not blocking landing,
+     * otherwise return the error message to be displayed on the PR
+    customChecks: [
       {
         rule: ({ pullRequest, buildStatuses, approvals, permissionLevel }) => true,
-        error: 'This check will always succeed',
       },
     ],
     */
--- a/config.example.js
+++ b/config.example.js
@@ -28,5 +28,19 @@ module.exports = {
     requireClosedTasks: true,
     requireGreenBuild: true,
     allowLandWhenAble: true,
+    /** What is provided to a custom rule:
+     *  {
+     *    pullRequest: BB.PullRequest -- see /src/bitbucket/types.d.ts
+     *    buildStatuses: BB.BuildStatus[] -- see /src/bitbucket/types.d.ts
+     *    approvals: string[] -- usernames of all real approvals
+     *    permissionLevel: "read" | "land" | "admin" -- permission level of the user requesting /can-land
+     *  }
+    rules: [
+      {
+        rule: ({ pullRequest, buildStatuses, approvals, permissionLevel }) => true,
+        error: 'This check will always succeed',
+      },
+    ],
+    */
   },
 };
--- a/src/bitbucket/BitbucketClient.ts
+++ b/src/bitbucket/BitbucketClient.ts
@@ -15,11 +15,11 @@ export class BitbucketClient {
 
   constructor(private config: Config) {}
 
-  async isAllowedToLand(pullRequestId: number) {
+  async isAllowedToLand(pullRequestId: number, permissionLevel: IPermissionMode) {
     const pullRequest: BB.PullRequest = await this.bitbucket.getPullRequest(pullRequestId);
     const buildStatuses = await this.bitbucket.getPullRequestBuildStatuses(pullRequestId);
     const author = pullRequest.author;
-    let approvals = getRealApprovals(
+    const approvals = getRealApprovals(
       pullRequest.approvals,
       author,
       this.config.prSettings.canApproveOwnPullRequest,
@@ -33,14 +33,14 @@ export class BitbucketClient {
       isApproved: approvals.length >= this.config.prSettings.requiredApprovals,
     };
 
+    const { prSettings } = this.config;
+    const errors: string[] = [];
+
     Logger.info('isAllowedToLand()', {
       pullRequestId,
       approvalChecks,
-      requirements: this.config.prSettings,
+      requirements: prSettings,
     });
-
-    const { prSettings } = this.config;
-    const errors: string[] = [];
 
     if (prSettings.requireClosedTasks && !approvalChecks.allTasksClosed) {
       errors.push(
@@ -58,7 +58,14 @@ export class BitbucketClient {
 
     if (!approvalChecks.isOpen) {
       errors.push('PR is already closed!');
+    } else if (prSettings.rules) {
+      prSettings.rules.forEach(({ rule, error }) => {
+        if (!rule({ pullRequest, buildStatuses, approvals, permissionLevel })) {
+          errors.push(error);
+        }
+      });
     }
+
     return {
       ...approvalChecks,
       errors,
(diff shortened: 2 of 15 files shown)


config.example.js
```diff
@@ -28,5 +28,19 @@ module.exports = {
...
+     *  {
+     *    pullRequest: BB.PullRequest -- see /src/bitbucket/types.d.ts
+     *    buildStatuses: BB.BuildStatus[] -- see /src/bitbucket/types.d.ts
+     *    approvals: string[] -- usernames of all real approvals
+     *    permissionLevel: "read" | "land" | "admin" -- permission level of the user requesting /can-land
+     *  }
+    rules: [
+      {
+        rule: ({ pullRequest, buildStatuses, approvals, permissionLevel }) => true,
+        error: 'This check will always succeed',
```
RajaBellebon: :)

src/bitbucket/BitbucketClient.ts
```diff
@@ -15,11 +15,11 @@ export class BitbucketClient {
 
   constructor(private config: Config) {}
 
-  async isAllowedToLand(pullRequestId: number) {
+  async isAllowedToLand(pullRequestId: number, permissionLevel: IPermissionMode) {
     const pullRequest: BB.PullRequest = await this.bitbucket.getPullRequest(pullRequestId);
     const buildStatuses = await this.bitbucket.getPullRequestBuildStatuses(pullRequestId);
     const author = pullRequest.author;
-    let approvals = getRealApprovals(
+    const approvals = getRealApprovals(
```
RajaBellebon: Are we sure we are not calling approvals elsewhere?

src/lib/Runner.ts
```diff
@@ -50,9 +50,16 @@ export class Runner {
         landRequest: landRequest.get(),
       });
 
+      const triggererUserMode = await permissionService.getPermissionForUser(
+        landRequest.triggererAaid,
+      );
+
       // TODO: Pass this commit in to isAllowed to land and make sure it hasnt changed
```
RajaBellebon: Is the TODO still valid?

src/lib/Runner.ts
```diff
@@ -50,9 +50,16 @@ export class Runner {
         landRequest: landRequest.get(),
       });
 
+      const triggererUserMode = await permissionService.getPermissionForUser(
```
RajaBellebon: hm I found the name of the function a bit weird - I tot understand but this is to get the permissionLevel right?

src/static/current-state/components/tabs/HistoryTab.tsx
```diff
@@ -7,14 +7,14 @@ import { QueueItemJoined } from '../QueueItem';
...
 
-export type HistoryItemsListProps = {
-  history: Array<HistoryItem>;
-  renderEmpty?: () => JSX.Element;
-  bitbucketBaseUrl: string;
-  items: number;
-  pageLen: number;
-  onPageChange: (page: number) => void;
-};
+// export type HistoryItemsListProps = {
```
RajaBellebon: Is that commented on purpose?

src/static/current-state/components/tabs/index.tsx
```diff
@@ -100,7 +100,7 @@ export class Tabs extends React.Component<TabsProps, TabsState> {
     }
 
     this.state = {
-      selected: isNaN(selected) ? 1 : selected,
+      selected: !selected || isNaN(selected) ? 1 : selected,
```
RajaBellebon: Have you checked the logic?

src/types.ts
```diff
@@ -12,12 +12,25 @@ export type RepoConfig = {
   uuid?: string;
 };
 
+type CustomRule = {
+  rule: (
+    pullRequestInfo: {
+      pullRequest: BB.PullRequest;
+      buildStatuses: BB.BuildStatus[];
+      approvals: string[];
```
RajaBellebon: jfmi - What do we store - the name? the number of approvals?

src/lib/Runner.ts
gardnerjack: Not sure, will have to ask Luke

src/lib/Runner.ts
gardnerjack: The terms 'user mode' and 'permission level' seem to be used interchangeably throughout the app, I can change it to make it clearer though.

src/static/current-state/components/tabs/HistoryTab.tsx
gardnerjack: It wasn't actually being used as it goes with some other code in that file that was already commented out. Not sure if it still needs to be there.

src/types.ts
gardnerjack: List of the usernames of the approvers

src/static/current-state/components/tabs/index.tsx
gardnerjack: I added this because isNaN on an empty string returns false, but seems like it's unnecessary since parseInt on an empty string is NaN.

src/bitbucket/BitbucketClient.ts
gardnerjack: Definitely not.

src/lib/Runner.ts
RajaBellebon: it might make it clearer

src/static/current-state/components/tabs/index.tsx
RajaBellebon: yeap

src/lib/Runner.ts
lukebatchelor: Yea, still valid.
The check here would be an edge case that would be super hard to hit.
Someone would have to start landing a PR, then push a change, hack in approvals, green builds, etc whilst landkid is still running, then recieve the green build for the previous state, all before the landkid build finished.

Would be pretty unlikely, but still technically doable

config.example.js
lukebatchelor: We'll want to document this properly at some point.
I know marco wants to start building on top of these already

src/bitbucket/BitbucketClient.ts
```diff
@@ -58,7 +58,18 @@ export class BitbucketClient {
...
       errors.push('PR is already closed!');
+    } else if (prSettings.rules) {
+      const pullRequestInfo = {
+        pullRequest,
+        buildStatuses,
+        approvals,
+        permissionLevel,
+      };
+      errors.push(
+        ...prSettings.rules.filter(({ rule }) => !rule(pullRequestInfo)).map(({ error }) => error),
```
lukebatchelor: This is a neat way of running them but is a bit tricky to parse imo.

Could we write this a bit clearer:

```
prSettings.customRules.forEach(({ rule, errorMsg }) => {
  const passesRule = rule(pullRequestInfo);
  if (!passesRule) errors.push(errorMsg);
});
```

config.example.js
lukebatchelor: A suggestion here (sorry, should have gotten feedback back to you sooner).

Just for future-proofings sake, could we:
* Make all these checks async (always await on them, I think a synchronous function should Just Work™ too)
* Remove the wrapping object and just have a function that returns either an empty string or an error message.

The second one is a bit weird, in that _normaly_ I'd be all for having these things named explicitly, especially with nice, simple default error messages. But in the future, we might also want to be able to return more complex error messages that might need to be quite dynamic. So I think returning them from the function would make sense? I did think about having it return `true` or an error message, but that gets a bit weird because strings a truthy and although you could detect it, it's just a bit weird...

I dunno, what do you think?

lukebatchelor:
Sweet, as discussed, we'll do a follow up PR here before releasing this

PR #40, opened 2019-12-04
User notes

gardnerjack:
New table `UserNote` that stores `aaid`, `note`, and `setByAaid`. This maintains a one-to-one mapping of `aaid` to a note, so when a note is created a new row is created, when the note is edited the existing row is updated, and when the note is removed the row is deleted.

When fetching the current state of the runner, notes are attached to the list of user objects if the requesting user is an admin. Notes and the associated controls are also only displayed to admins.


src/lib/PermissionService.ts
```diff
@@ -36,6 +36,37 @@ class PermissionService {
...
+        where: {
+          aaid,
+        },
+      },
+    ))[0] > 0 || // Check if a row was updated, otherwise create a new row
+      (await UserNote.create<UserNote>({
+        aaid,
+        note,
+        setByAaid: setter.aaid,
+      }));
```
lukebatchelor: This feels a little messy and hard to read. Could we do something like:

```
const numNotesForUser = await UserNote.count({ where: { aaid }});
if (numNotesForUser === ) { // create the first note
  await UserNote.create<UserNote>({ aaid, note, setByAaid: setter.aaid });
} else { // update the existing one
  await UserNote.update({ note, setByAaid: setter.aaid }, { where: aaid });
}
```

src/lib/PermissionService.ts
```diff
@@ -36,6 +36,37 @@ class PermissionService {
...
+  };
+
+  removeUserNote = async (aaid: string): Promise<void> => {
```
lukebatchelor: Absolutely not worried, but do you think we actually need to destroy them? Or just mark them blank?
Definitely not worth changing, just wondering if there was a reason for it

src/static/current-state/components/tabs/UsersList.tsx
```diff
@@ -0,0 +1,152 @@
+import * as React from 'react';
```
lukebatchelor: This is super clean, nice!

src/lib/PermissionService.ts
gardnerjack: Just in the interest of maintaining a specific pattern for this table. Where the row exists if the note exists, and there is one row/note per user.

PR #41, opened 2019-12-04
Cleanup

gardnerjack:
Just cleaning up a bunch of random stuff for more concise/consistent code. Not sure if there are any other changes that should be in here.

I also realised that since we haven't deployed the banner message feature we can get rid of the migration for creating the `MessageStateTransition` table.


lukebatchelor:
Nice, this is awesome!

PR #45, opened 2019-12-17
One row table pattern for Landkid states

gardnerjack:
New tables `PauseState` and `BannerMessageState` with one row (or none) that represents the existing state both replace the previous pattern of state transition tables.


src/lib/Runner.ts
```diff
@@ -214,71 +214,42 @@ export class Runner {
   };
 
   pause = async (reason: string, user: ISessionUser) => {
-    await PauseStateTransition.create<PauseStateTransition>({
-      paused: true,
-      reason,
+    this.unpause();
```
lukebatchelor: This should have an await on it

src/lib/Runner.ts
```diff
@@ -214,71 +214,42 @@ export class Runner {
...
   };
 
-  sendBannerMessage = async (
+  addBannerMessage = async (
     message: string,
     messageType: IMessageState['messageType'],
     user: ISessionUser,
   ) => {
-    await MessageStateTransition.create<MessageStateTransition>({
+    this.removeBannerMessage();
```
lukebatchelor: await here too

PR #60, opened 2020-02-26
Endpoint for moving requests to the top of the queue

gardnerjack:
Just in case there's ever a scenario where it's needed, only admins can move requests to the top.

Also if you want to accept bribes that's fine with me.


src/lib/Runner.ts
```diff
@@ -333,6 +333,21 @@ export class Runner {
     return true;
   };
 
+  moveRequestToTopOfQueue = async (requestId: string, user: ISessionUser): Promise<boolean> => {
+    const landRequestStatus = await this.queue.maybeGetStatusForQueuedRequestById(requestId);
+    if (!landRequestStatus) return false;
+
+    const requestAtTop = (await this.getQueue())[0];
```
lukebatchelor: Hmmm.. What happens if we call this and theres nothing there?
I guess it could blow up, but wouldn't really break anything?

Might be nice to guard either way though

gardnerjack: True, will push quick fix

PR #62, opened 2020-02-26
Check if running builds have any failed dependencies before adding them as a dep

gardnerjack:
I think this might fix the maxConcurrentBuilds issue as well


lukebatchelor:
Sweet, nice clean up too.

PR #66, opened 2020-03-17
Improved Logging

gardnerjack:
Added new `verbose` logging level for local development.

Added namespaces to all logs, which can be filtered with:
- locally with the `LOG_NAMESPACES` env var
  - `LOG_NAMESPACES=routes:api:* yarn dev` will only shows logs from `src/routes/api`
- in splunk with `namespace` variable
  - `namespace=routes:api:*`

Added basic indicators to as many logs as possible: `requestId`, `pullRequestId`, `buildId`, and similar. Allows these variables to be used in Splunk to follow the flow of requests, pull requests, builds, etc.

### DIFF
--- a/src/bitbucket/BitbucketAPI.ts
+++ b/src/bitbucket/BitbucketAPI.ts
@@ -5,6 +5,7 @@ import * as pRetry from 'p-retry';
 import { RepoConfig } from '../types';
 import { Logger } from '../lib/Logger';
 import { bitbucketAuthenticator, axiosPostConfig } from './BitbucketAuthenticator';
+import { LandRequest } from '../db';
 
 const baseApiUrl = 'https://api.bitbucket.org/2.0/repositories';
 
@@ -13,7 +14,12 @@ export class BitbucketAPI {
 
   constructor(private config: RepoConfig) {}
 
-  mergePullRequest = async (pullRequestId: number, targetBranch: string) => {
+  mergePullRequest = async (landRequest: LandRequest) => {
+    const {
+      id: requestId,
+      pullRequestId,
+      pullRequest: { targetBranch },
+    } = landRequest;
     const endpoint = `${this.apiBaseUrl}/pullrequests/${pullRequestId}/merge`;
     const message = `pull request #${pullRequestId} merged by Landkid after a successful build rebased on ${targetBranch}`;
     const data = {
@@ -25,7 +31,9 @@ export class BitbucketAPI {
     Logger.info('Merging pull request', {
       namespace: 'bitbucket:api:mergePullRequest',
       pullRequestId,
-      request: { endpoint, ...data },
+      requestId,
+      targetBranch,
+      postRequest: { endpoint, ...data },
     });
     // This is just defining the function that we will retry
     const attemptMerge = async () =>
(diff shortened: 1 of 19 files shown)


gardnerjack:
[New logging system](https://trello.com/c/3uN6o1r0/31-new-logging-system)

src/bitbucket/BitbucketAPI.ts
```diff
@@ -51,12 +64,25 @@ export class BitbucketAPI {
...
+          namespace: 'bitbucket:api:mergePullRequest',
+          requestId,
+          pullRequestId,
+        }),
+      )
       .catch(err => {
-        Logger.error('Unable to merge pull request', { err, pullRequestId });
+        Logger.error('Unable to merge pull request', {
+          namespace: 'bitbucket:api:mergePullRequest',
+          err,
```
lukebatchelor: I recall some sort of issue before with loggin err objects... 
Do you remember what that was? It was something super weird....

lukebatchelor: Im pretty sure this is fine, but trying to recall

src/lib/Logger.ts
```diff
@@ -1,18 +1,29 @@
 import * as winston from 'winston';
+import { isMatch } from 'micromatch';
 
 process.stdout.isTTY = true;
 
 const ProdLogger = winston.createLogger({
-  level: 'info',
+  level: 'http',
```
lukebatchelor: This isnt going to remove all info messages from prod is it?

src/lib/Runner.ts
```diff
@@ -54,7 +54,15 @@ export class Runner {
     const runningTargetingSameBranch = running.filter(
       build => build.request.pullRequest.targetBranch === landRequest.pullRequest.targetBranch,
     );
-    if (runningTargetingSameBranch.length >= this.getMaxConcurrentBuilds()) return false;
+    const maxConcurrentBuilds = this.getMaxConcurrentBuilds();
+    if (runningTargetingSameBranch.length >= maxConcurrentBuilds) {
+      Logger.verbose('No concurrent build slots left', {
```
lukebatchelor: I think all the logging in here is going to be the biggest win

src/lib/Logger.ts
```diff
@@ -1,18 +1,29 @@
...
   format: winston.format.json(),
   transports: [new winston.transports.Console()],
 });
-const DevelopmentLogger = winston.createLogger({
-  level: 'info',
-  format: winston.format.combine(winston.format.colorize(), winston.format.simple()),
+
+const DevLogger = winston.createLogger({
+  level: 'verbose',
```
lukebatchelor: So we'll only get all those extra logs in dev?
Maybe we'll need to move a few key ones out of verbose and into info then?

Just so that we're getting enough info in our prod logs...
At the same time, I dont want to dump too much, not sure what the limits are, but people definitely hit them

lukebatchelor:
Sweet, my only worry is that 'http' level, just that we're not logging even less in prod.
The rest is a balancing act of wanting to log enough in prod... Any of those you feel like need to be moved?
I'd be tempted to move all the verbose ones that happen in non-often called flows to info. i.e, the ones in runner where we're moving state around. That shouldn't be called _too_ often I believe, so maybe they can be info's instead? The'd give us a lot of useful information when debuggin in prod

src/lib/Logger.ts
gardnerjack: This means that prod logs all levels `http` and below. 
Levels are defined as:
```
{ 
  error: 0, 
  warn: 1, 
  info: 2, 
  http: 3,
  verbose: 4, 
  debug: 5, 
  silly: 6 
}
```

src/lib/Logger.ts
gardnerjack: I'll have a look through the ones I've made `verbose` and see if any could be useful in a prod scenario

PR #67, opened 2020-03-31
Custom Warnings + Allow use of Axios in custom rules

gardnerjack:
Added new category of custom rules, are displayed on the landkid PR view if they fail but will not prevent landing.
Also allow use of axios in custom rules.

### DIFF
--- a/src/bitbucket/BitbucketClient.ts
+++ b/src/bitbucket/BitbucketClient.ts
@@ -37,6 +37,7 @@ export class BitbucketClient {
 
     const { prSettings } = this.config;
     const errors: string[] = [];
+    const warnings: string[] = [];
 
     Logger.info('Pull request approval checks', {
       namespace: 'bitbucket:client:isAllowedToLand',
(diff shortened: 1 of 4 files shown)


src/bitbucket/BitbucketClient.ts
```diff
@@ -60,22 +62,31 @@ export class BitbucketClient {
...
         approvals,
         permissionLevel,
       };
-      for (const { rule } of prSettings.customChecks) {
-        const passesRule = await rule(pullRequestInfo);
-        if (typeof passesRule === 'string') errors.push(passesRule);
+      if (prSettings.customChecks) {
+        for (const { rule } of prSettings.customChecks) {
+          const passesRule = await rule(pullRequestInfo, axios);
+          if (typeof passesRule === 'string') errors.push(passesRule);
```
lukebatchelor, line 75: Hmmm.. I get why you added the axios injection, but it still feels a bit iffy to me...
Lets go with it for now and see if we need to change it later I guess.

I think in the end, if you config needs external packages, you should just have them installed. But lets go with this for now

PR #68, opened 2020-05-08
Set inner html on rule rendering

gardnerjack:
Not sure about this one. I want to be able to format the output of rules, but I think this probably could allow for some dodgy stuff. Thoughts?

Use case: https://bitbucket.org/atlassian/atlassian-frontend-landkid-deployment/pull-requests/31

Also install `httplease-asap` as a devDep for locally testing the `package-ownership` rule.


lukebatchelor:
Sweet, yea. Deffs get the concern. It's not user generated in any way, so I'm happy with it.

PR #76, opened 2020-07-06
More logging

gardnerjack:
Action items from https://hello.atlassian.net/wiki/spaces/~904291390/pages/767240910/Weirdness

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "pretest": "node tools/test-local.js",
     "test": "yarn test:staging",
     "posttest": "node tools/revert-test-local.js",
-    "test:staging": "yarn run cypress run",
+    "test:staging": "yarn run cypress run --reporter junit",
     "format": "prettier --write src/**/*.ts src/**/*.tsx",
     "build": "yarn run clean && yarn build:server && yarn build:ui",
     "build:server": "tsc",
--- a/src/bitbucket/BitbucketAPI.ts
+++ b/src/bitbucket/BitbucketAPI.ts
@@ -5,7 +5,7 @@ import * as pRetry from 'p-retry';
 import { RepoConfig } from '../types';
 import { Logger } from '../lib/Logger';
 import { bitbucketAuthenticator, axiosPostConfig } from './BitbucketAuthenticator';
-import { LandRequest } from '../db';
+import { LandRequestStatus } from '../db';
 
 const baseApiUrl = 'https://api.bitbucket.org/2.0/repositories';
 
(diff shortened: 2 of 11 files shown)


gardnerjack:
**Deployment:** https://bitbucket.org/atlassian/atlassian-frontend-landkid-deployment/addon/pipelines/home#!/results/241

package.json
```diff
@@ -20,7 +20,7 @@
     "pretest": "node tools/test-local.js",
     "test": "yarn test:staging",
     "posttest": "node tools/revert-test-local.js",
-    "test:staging": "yarn run cypress run",
+    "test:staging": "yarn run cypress run --reporter junit",
```
RajaBellebon, line 23: small nit

do we need to do `yarn run cypresss ...`

You could do

```
   "test:watch": "cypress open",
    "test:chrome": "cypress run --browser chrome  --reporter junit",
    "test:firefox": "cypress run --browser firefox  --reporter junit",
    "test": "yarn test:chrome && test:firefox"
```

No need to do that in this PR - and yes cross browser is supported since version 4

src/bitbucket/BitbucketAPI.ts
```diff
@@ -31,8 +33,9 @@ export class BitbucketAPI {
     Logger.info('Merging pull request', {
       namespace: 'bitbucket:api:mergePullRequest',
       pullRequestId,
-      requestId,
+      landRequestId,
       targetBranch,
+      landRequestStatus,
```
RajaBellebon, line 38: so we will display the LandRequestStatus sent to Splunk
* awaiting merge
* ...

src/bitbucket/BitbucketClient.ts
```diff
@@ -90,16 +90,16 @@ export class BitbucketClient {
     };
   }
 
-  createLandBuild(requestId: string, commit: string, variables: PipelinesVariables) {
-    return this.pipelines.createLandBuild(requestId, commit, variables);
+  createLandBuild(requestId: string, commit: string, variables: PipelinesVariables, lockId: Date) {
```
RajaBellebon, line 93: jfmi why `lockId` is required for create but not for stop?

RajaBellebon: and what will be the purpose of it?

src/static/bitbucket/components/App.tsx
```diff
@@ -31,7 +31,7 @@ export class App extends React.Component {
 
   messageEmoji = {
     default: '📢',
-    warning: '⚠',
+    warning: '⚠️',
```
RajaBellebon, line 34: ?

gardnerjack: I've got no idea what happened here, I think I accidentally changed it in a previous PR. It looks the same but has a different unicode I think.

src/bitbucket/BitbucketClient.ts
gardnerjack: `lockId` is used in every function that is called from `next`. Since `next` transitions a request from queued to running it creates pipelines but doesn't stop them. I explain the purpose of the `lockId` at the bottom of the investigation page, but essentially it is to see what `next` call the functions are coming from.

src/bitbucket/BitbucketAPI.ts
gardnerjack: Exactly, when just logging `landRequest` we couldn't see the current state.

package.json
gardnerjack: True, could remove the `yarn run` bit, but not sure the value of running the tests on different browsers for now.

PR #77, opened 2020-07-13
New BB merge API handling

gardnerjack:
Added handling for new `202` response from BB merge API:
- https://staging.bb-inf.net/bitbucket/api-specifications/src/master/v20/pr_merge_async.md
- https://developer.atlassian.com/bitbucket/api/2/reference/resource/repositories/%7Bworkspace%7D/%7Brepo_slug%7D/pullrequests/%7Bpull_request_id%7D/merge

Also rewrote how the merge function works based on discussion surrounding the topic of how Landkid merges PRs:
- https://product-fabric.atlassian.net/l/c/BdCCRjgv

@lukebatchelor not sure how to test this since we don't have good (any) unit testing, but maybe you can catch some issues with the code that I've missed so far.


src/bitbucket/BitbucketAPI.ts
```diff
@@ -1,6 +1,6 @@
-import axios, { AxiosError } from 'axios';
+import axios, { AxiosResponse } from 'axios';
 import * as jwtTools from 'atlassian-jwt';
-import * as pRetry from 'p-retry';
+import delay from 'delay';
```
RajaBellebon, line 3: Can you elaborate a bit this change of lib?

src/bitbucket/BitbucketAPI.ts
```diff
@@ -30,67 +30,186 @@ export class BitbucketAPI {
...
-
-      // NOTE: Do **NOT** log the whole failure object here, it will have the axios config in it
-      // which contains auth credentials
-
-      Logger.error('Merge attempt failed', {
-        namespace: 'bitbucket:api:mergePullRequest:onFailedAttept',
+
+    // Polls result of merge task if the merge takes more than 28 seconds
+    // API returns 202 if this is required
+    const pollTaskResult = async (pollUrl: string): Promise<any> => {
```
RajaBellebon, line 44: You can add simple unit test to your function with API calls using axios - check the bundle size add-on test

src/bitbucket/BitbucketAPI.ts
```diff
@@ -30,67 +30,186 @@ export class BitbucketAPI {
...
+          Logger.info('202: delaying and polling again', {
+            // TODO: REMOVE AFTER TESTING
+            namespace: 'bitbucket:api:mergePullRequest:attemptMerge',
+            pullRequestId,
+            landRequestId,
+            pollUrl,
+            res,
+          });
+          // poll every 3 seconds
+          await delay(3000);
```
RajaBellebon, line 71: Is that the outcome of the discussion with Rani?
If yes 👍

src/bitbucket/BitbucketClient.ts
```diff
@@ -95,11 +95,11 @@ export class BitbucketClient {
   }
 
   async stopLandBuild(buildId: number, lockId?: Date) {
-    return await this.pipelines.stopLandBuild(buildId, lockId);
+    return this.pipelines.stopLandBuild(buildId, lockId);
```
RajaBellebon, line 98: oh I don;t remember if eslint is setup In Landkid but it would have been captured - nice

tsconfig.json
```diff
@@ -1,24 +1,25 @@
...
-		"strictNullChecks": true,
-		"noUnusedLocals": true,
-		"noImplicitThis": true,
-		"noImplicitAny": true
-	},
-	"exclude": [
-		"node_modules",
-		"public"
-	]
+    "compilerOptions": {
```
RajaBellebon, line 2: What changed here?

src/bitbucket/BitbucketAPI.ts
RajaBellebon: I think worth investigating a bit of time in unit test & think about the edge cases - for example when we have concurrency set

tsconfig.json
gardnerjack: Just formatting

src/bitbucket/BitbucketAPI.ts
gardnerjack: Have replaced `p-retry` with doing it manually because I tried to reduce the cases in which we retry the API, and `p-retry` has semi-opinionated so was easier to do it manually.

`delay` is just used to poll the merge task every 3 seconds.

src/bitbucket/BitbucketAPI.ts
gardnerjack: Yeah I'll look into writing a couple unit tests. I tried to do it with cypress but was having some difficulties with mocking everything and felt like it was going to become a time sink. Might try with jest because I'm more familiar with writing unit tests that way.

RajaBellebon: There is a [cypress-testing-libraries](https://github.com/testing-library/cypress-testing-library) but not sure how good it is - can go with Jest - easy to setup and write

lukebatchelor:
Awesome, extra logging looks great and is only called on merges so shouldn't cause too much load at all. Nice on

PR #81, opened 2020-07-24
Add status-transition lock to checkWaitingLandRequests

gardnerjack:
`checkWaitingLandRequests` has a similar functionality to `next`, it iterates over a list of requests and performs status transition actions. We lock `next` so there are no conflicts, we should also do that here.

[see investigation here](https://hello.atlassian.net/wiki/spaces/~904291390/pages/767240910/Weirdness)

### DIFF
--- a/src/lib/Runner.ts
+++ b/src/lib/Runner.ts
@@ -222,7 +222,7 @@ export class Runner {
 
   // Next must always return early if ever doing a single state transition
   next = async () => {
-    await withLock('Runner:next', async (lockId: Date) => {
+    await withLock('status-transition', async (lockId: Date) => {
       const queue = await this.queue.getQueue();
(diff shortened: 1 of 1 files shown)


src/lib/Runner.ts
```diff
@@ -415,6 +415,7 @@ export class Runner {
     this.checkWaitingLandRequests();
   };
 
+  // TODO: CHECK FOR LOCK
```
lukebatchelor, line 418: This one still todo?

gardnerjack: oops nope, that is only called by checkWaitingLandRequests so it's fine

gardnerjack:
Manually tested, transition is successfully locked when adding request to land when able, and next is not called until after it is finished.

PR #82, opened 2020-08-27
[AFP-2121] Configure Permissions Message

thebriando:
Adds a `permissionsMessage` field to the config to configure the message when a user views the queue tab or history tab and does not have admin access.

### DIFF
--- a/config.example.js
+++ b/config.example.js
@@ -25,6 +25,7 @@ module.exports = {
     },
   },
   maxConcurrentBuilds: 3,
+  permissionsMessage: 'Contact an admin for permission to view this information',
   prSettings: {
     requiredApprovals: 1,
     canApproveOwnPullRequest: true,
--- a/config.example.js
+++ b/config.example.js
@@ -25,7 +25,6 @@ module.exports = {
     },
   },
   maxConcurrentBuilds: 3,
-  permissionsMessage: 'Contact an admin for permission to view this information',
   prSettings: {
     requiredApprovals: 1,
     canApproveOwnPullRequest: true,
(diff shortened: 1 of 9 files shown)


gardnerjack:
Awesome stuff, don't worry about the integration tests failing. You don't actually need to run them in this case because we have this weird set up where you need to deploy the changes first before running the tests (so it's pointless to do it here before doing that).
I'll take you through that process when we deploy this change :)

lukebatchelor (approved):
Awesome!!! Looks great. 

For our message when you deploy, let's have it be something like "please ping !disturbed in #atlassian-frontend if you require access to land"

gardnerjack:
I reckon that we should have a default permissions message, which would reduce upgrade pain. We don't really have a proper way of doing this (although we probably should), so maybe in `Config.ts` include something like this:

```
if (!config.permissionsMessage) {
  config.permissionsMessage = 'Contact an admin...
}
```

Then you would have to update some of the types so that you can assume existence of the message.

gardnerjack (changes requested):
I reckon that we should have a default permissions message, which would reduce upgrade pain. We don't really have a proper way of doing this (although we probably should), so maybe in `Config.ts` include something like this:

```
if (!config.permissionsMessage) {
  config.permissionsMessage = 'Contact an admin...
}
```

Then you would have to update some of the types so that you can assume existence of the message.

gardnerjack:
To add on to my comment above, I think it would be great to actually have some proper config validation with defaults and stuff. I know [convict](https://github.com/mozilla/node-convict/tree/master/packages/convict) provides this pretty well. But [joi](https://github.com/sideway/joi) is a dev dep for testing purposes so maybe that will offer a good option as well. Don't need to do it for this but just wanted to document my thoughts.

gardnerjack (approved):

PR #84, opened 2020-09-09
AFP-2125 Landkid Blame

thebriando:
- Adds `Landed by:` field to queue items on the frontend
- Refactors status reason to use the user's `displayName` if it exists, and the `aaid` if it doesn't

### DIFF
--- a/src/lib/Runner.ts
+++ b/src/lib/Runner.ts
@@ -469,7 +469,7 @@ export class Runner {
 
     await landRequestStatus.request.setStatus(
       'queued',
-      `moved to the top of the queue by user ${user.aaid}`,
+      `moved to the top of the queue by user ${user.displayName || user.aaid}`,
       topDate,
     );
     return true;
--- a/src/lib/Runner.ts
+++ b/src/lib/Runner.ts
@@ -214,8 +214,7 @@ export class Runner {
         pullRequestId,
         lockId,
       });
-      const user = await this.client.getUser(landRequest.triggererAaid);
-      return landRequest.setStatus('success', `Landed by ${user.displayName || user.aaid}`);
+      return landRequest.setStatus('success');
     } catch (err) {
       return landRequest.setStatus('fail', 'Unable to merge pull request');
     }
@@ -250,7 +249,7 @@ export class Runner {
           await landRequest.update({ dependsOn: null });
           await this.client.stopLandBuild(landRequest.buildId, lockId);
           const user = await this.client.getUser(landRequest.triggererAaid);
-          return landRequest.setStatus('queued', `queued by ${user.displayName || user.aaid}`);
+          return landRequest.setStatus('queued', `Queued by ${user.displayName || user.aaid}`);
         }
         if (landRequestStatus.state === 'awaiting-merge') {
           const didChangeState = await this.moveFromAwaitingMerge(landRequestStatus, lockId);
@@ -327,7 +326,7 @@ export class Runner {
 
     await landRequestStatus.request.setStatus(
       'aborted',
-      `Cancelled by user "${user.displayName || user.aaid}"`,
+      `Cancelled by user ${user.displayName || user.aaid}`,
     );
     if (landRequestStatus.request.buildId) {
       return this.client.stopLandBuild(landRequestStatus.request.buildId);
@@ -405,7 +404,7 @@ export class Runner {
     if (await this.getPauseState()) return;
     const request = await this.createRequestFromOptions(landRequestOptions);
     const user = await this.client.getUser(request.triggererAaid);
-    return request.setStatus('queued', `queued by ${user.displayName || user.aaid}`);
+    return request.setStatus('queued', `Queued by ${user.displayName || user.aaid}`);
   };
 
   addToWaitingToLand = async (landRequestOptions: LandRequestOptions) => {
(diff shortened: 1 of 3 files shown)


src/static/current-state/components/QueueItem.tsx
```diff
@@ -259,11 +258,22 @@ export class QueueItem extends React.Component<QueueItemProps, QueueItemState> {
             </StatusItem>
           ))}
         </div>
-        {status.reason ? (
+        {status.reason && status.state !== 'success' ? (
```
gardnerjack, line 261: What was the reason behind this change?

src/static/current-state/components/QueueItem.tsx
```diff
@@ -259,11 +258,22 @@ export class QueueItem extends React.Component<QueueItemProps, QueueItemState> {
...
+          <StatusItem title="Landed by:">
+            <Lozenge>
+              <User aaid={status.request.triggererAaid}>
+                {user => {
+                  return user.displayName;
+                }}
+              </User>
+            </Lozenge>
+          </StatusItem>
+        ) : null}
```
gardnerjack, line 276: I do like this, but the reason section will display this info when a request has been queued, perhaps you could change the status state check on the reason section to be `status.state !== 'queued'`?

src/static/current-state/components/QueueItem.tsx
thebriando: My reasoning behind this was because the "Landed by:" field I added has the name of the triggerer so it wouldn't make sense to display the reason too in this case since the reason would contain the name too if the state is `success`.

src/static/current-state/components/QueueItem.tsx
thebriando: Hmm, I added a reason to a `success` status too, should I take this out since it'll show up twice? Or change the check so it's `status.state !== 'queued' && status.state !== 'success'`?

gardnerjack: Ahh didn't see that one, I would remove the `Landed by` reason from the `success` status transition, and change that check just to `status.state !== 'queued'`.

src/lib/Runner.ts
```diff
@@ -248,8 +249,8 @@ export class Runner {
           await landRequest.setStatus('fail', failReason);
           await landRequest.update({ dependsOn: null });
           await this.client.stopLandBuild(landRequest.buildId, lockId);
-          // await landRequest.request.save();
-          return landRequest.setStatus('queued');
+          const user = await this.client.getUser(landRequest.triggererAaid);
+          return landRequest.setStatus('queued', `queued by ${user.displayName || user.aaid}`);
```
gardnerjack, line 253: nit: capitalise Queued

src/lib/Runner.ts
```diff
@@ -326,7 +327,7 @@ export class Runner {
 
     await landRequestStatus.request.setStatus(
       'aborted',
-      `Cancelled by user "${user.aaid}" (${user.displayName})`,
+      `Cancelled by user "${user.displayName || user.aaid}"`,
```
gardnerjack, line 330: nit: remove quotations around name for consistency

src/lib/Runner.ts
```diff
@@ -403,7 +404,8 @@ export class Runner {
     // TODO: Ensure no land request is pending for this PR
     if (await this.getPauseState()) return;
     const request = await this.createRequestFromOptions(landRequestOptions);
-    await request.setStatus('queued');
+    const user = await this.client.getUser(request.triggererAaid);
+    return request.setStatus('queued', `queued by ${user.displayName || user.aaid}`);
```
gardnerjack, line 408: another capitalisation here

src/lib/Runner.ts
```diff
@@ -427,7 +429,8 @@ export class Runner {
       const status = await request.getStatus();
       if (status && status.state !== 'will-queue-when-ready') continue;
 
-      await request.setStatus('queued');
+      const user = await this.client.getUser(request.triggererAaid);
+      await request.setStatus('queued', `queued by ${user.displayName || user.aaid}`);
```
gardnerjack, line 433: and here :)

src/lib/Runner.ts
```diff
@@ -444,7 +447,7 @@ export class Runner {
     const displayName = user.displayName || user.aaid;
     await landRequestStatus.request.setStatus(
       'aborted',
-      `Removed from queue by user "${displayName}`,
+      `Removed from queue by user "${displayName}"`,
```
gardnerjack, line 450: also remove quotations here

src/lib/Runner.ts
```diff
@@ -466,7 +469,7 @@ export class Runner {
 
     await landRequestStatus.request.setStatus(
       'queued',
-      `moved to the top of the queue by user ${user.aaid}`,
+      `moved to the top of the queue by user ${user.displayName || user.aaid}`,
```
gardnerjack, line 472: and a capital M please

gardnerjack (changes requested):

gardnerjack:
Sorry for all of the above, I'm a sucker for consistency ahaha

thebriando:
No problem, should be fixed now 😄  @gardnerjack

src/static/current-state/components/QueueItem.tsx
```diff
@@ -259,11 +258,22 @@ export class QueueItem extends React.Component<QueueItemProps, QueueItemState> {
             </StatusItem>
           ))}
         </div>
-        {status.reason ? (
+        {status.reason && status.reason !== 'queued' ? (
```
gardnerjack, line 261: This should be `status.state !== 'queued'`

gardnerjack:
Awesome work 💯, just that one small thing then we're good to go. If you could raise a deployment PR once this is merged that would be great, we can deploy that to dev and do last minute testing but I'm sure it'll work just fine.

gardnerjack (approved):

PR #85, opened 2020-09-14
Add authentication and typing to polling merge status

gardnerjack:
https://product-fabric.atlassian.net/browse/AFP-2402

### DIFF
--- a/src/bitbucket/types.d.ts
+++ b/src/bitbucket/types.d.ts
@@ -44,6 +44,15 @@ declare namespace BB {
     url: string;
   };
 
+  type MergeStatusResponse =
+    | {
+        task_status: 'PENDING';
+      }
+    | {
+        task_status: 'SUCCESS';
+        merge_result: any;
+      };
+
   type PullRequest = {
     pullRequestId: number;
     title: string;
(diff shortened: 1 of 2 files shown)


src/bitbucket/types.d.ts
```diff
@@ -44,6 +44,15 @@ declare namespace BB {
     url: string;
   };
 
+  type MergeStatusResponse =
+    | {
+        task_status: 'PENDING';
+      }
+    | {
+        task_status: 'SUCCESS';
+        merge_result: any;
```
RajaBellebon, line 53: I guess this is fine for now ?

RajaBellebon (approved):

src/bitbucket/types.d.ts
gardnerjack: Yeah was about to add an explanation. We don't currently make use of the merge_result for any type of merge so didn't even really need to include this field but thought I might as well so we know the general shape and can add typing to this if required later.

RajaBellebon: if we have a ticket or task later for this or just a comment might be worth (not in this PR if you don't want)

PR #87, opened 2020-10-26
Self-dependencies

gardnerjack:


src/lib/Runner.ts
```diff
@@ -249,21 +252,26 @@ export class Runner {
...
-          if (movedState) return this.next();
+          return !!didChangeState;
         }
         // otherwise, we must just be running, nothing to do here
       }
+      return false;
     });
+    if (runNextAgain) {
+      await this.next();
+    }
```
gardnerjack, line 274: Essentially I changed this because I don't think it was causing a lot of the problems to call `next` from inside the locked `next` call. I'm not sure what the specific details are with `awaiting` a recursive call though...

src/lib/Runner.ts
```diff
@@ -53,7 +53,10 @@ export class Runner {
     const landRequest = landRequestStatus.request;
     const running = await this.getRunning();
     const runningTargetingSameBranch = running.filter(
-      build => build.request.pullRequest.targetBranch === landRequest.pullRequest.targetBranch,
+      build =>
+        build.request.pullRequest.targetBranch === landRequest.pullRequest.targetBranch &&
+        // Failsafe to prevent self-dependencies
+        build.request.pullRequestId !== landRequest.pullRequestId,
```
gardnerjack, line 59: Adding this failsafe just in case as well

lukebatchelor:
Sweet, looks good pending test in staging

lukebatchelor:
Sweet, had a second look and I think the moving out the .next call sounds like a good assumption.

lukebatchelor (approved):

PR #89, opened 2020-11-10
new IsAllowedToLand function that includes an existing request check

gardnerjack:
- changed `isAllowedToLand` -> `isAllowedToMerge`
- created new `isAllowedToLand` in the runner that calls `isAllowedToMerge` then does a check for an existing request
  - Accepts a function that returns a list of land requests so that the existing request can be searched for in different queues

### DIFF
--- a/src/lib/Runner.ts
+++ b/src/lib/Runner.ts
@@ -49,6 +49,12 @@ export class Runner {
     return this.queue.getRunning();
   };
 
+  getWaitingAndQueued = async () => {
+    const queued = await this.getQueue();
+    const waiting = await this.queue.getStatusesForWaitingRequests();
+    return [...queued, ...waiting];
+  };
+
   moveFromQueueToRunning = async (landRequestStatus: LandRequestStatus, lockId: Date) => {
     const landRequest = landRequestStatus.request;
     const running = await this.getRunning();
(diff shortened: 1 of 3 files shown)


gardnerjack:
Deploying to dev for testing: https://bitbucket.org/atlassian/atlassian-frontend-landkid-deployment/addon/pipelines/home#!/results/313

src/lib/Runner.ts
```diff
@@ -49,14 +49,17 @@ export class Runner {
...
+    const waiting = await this.queue.getStatusesForWaitingRequests();
+    return [...queued, ...waiting];
+  };
+
   moveFromQueueToRunning = async (landRequestStatus: LandRequestStatus, lockId: Date) => {
     const landRequest = landRequestStatus.request;
     const running = await this.getRunning();
     const runningTargetingSameBranch = running.filter(
-      build =>
-        build.request.pullRequest.targetBranch === landRequest.pullRequest.targetBranch &&
```
lukebatchelor, line 57: So, we've removed this, but it's not taken care of by `isAllowedToMerge` is that correct?

src/lib/Runner.ts
```diff
@@ -120,8 +124,12 @@ export class Runner {
       return landRequest.setStatus('fail', 'Unable to land due to failed land checks');
     }
 
+    const runningExceptSelf = runningTargetingSameBranch.filter(
+      // Failsafe to prevent self-dependencies
+      build => build.request.pullRequestId !== landRequest.pullRequestId,
```
lukebatchelor, line 129: Oh, nuh, moved down here. Yup

lukebatchelor:
Sweet, lgtm

PR #90, opened 2020-11-30
Bump to node 12.20.0

RajaBellebon:
### Description

Bumping from node 8 to node 12 to avoid security issue.

Added node to `engines` && deduplicate `fsevents`.

If you are on MacOS and you have fsevents not installed or installed (as for most of the packages it is an optional dependencies) but with a a version that does not support Node 12 and you use webpack, it will lead to performance issues + a lot of warning when running `yarn`.

_Related info:_

* https://til.codes/fix-for-100-cpu-usage-by-nodejs/
* https://github.com/fsevents/fsevents/issues/278

I replaced the `circleci/node` by `cimg/node` as per the documentation https://circleci.com/developer/images/image/cimg/node

I had to add the version of the `setup_remote_docker` due to this issue https://support.circleci.com/hc/en-us/articles/360050934711

### Testing in progress

* Locally

- [x] yarn
- [x] yarn jest
- [x] yarn build

* CI
- [x] CI

* End to End
- [x] deploy to ddev and run integration tests -> [pipeline link](https://app.circleci.com/pipelines/github/atlassian/landkid/261/workflows/b0463566-5c62-492f-97e5-5fd55aea9832) link 
- [x] Manual check for errors / runtime errors on dev -> [splunk link](https://splunk.paas-inf.net/en-US/app/search/search?q=search%20eventtype%3D%22micros_atlassian-frontend-landkid%22%20(application%20m.t%3Dapplication%20NOT%20micros_container%3Dplatform-*)%20(adev%20env%3Dadev)%20level%3Derror&display.page.search.mode=smart&dispatch.sample_ratio=1&earliest=-24h%40h&latest=now&sid=1606973046.19434_8A2D8EF0-881B-45D1-84B2-1B0BF4E3958A)


RajaBellebon:
@itsjackgardner . @lukebatchelor do I need to test something else?

Also are you fine with node 12 - I could try to bump to node 14?

RajaBellebon:
@itsjackgardner after looking again to the dev PRs, I saw that
Screen Shot 2020-12-03 at 21 55 51
why the re-entry success failed? is that expected?

Furthermore, digging more in the splunk logs, I found an error
```
level":"error","m":{"sv":"BB_BUILD_319","si":"atlassian-frontend-landkid","di":"a0mqhl2ptsn4rm01","g":"WebServer","t":"application"},"message":"Error failed while in lock for \"status-transition\"","micros_container":"microservice","namespace":"lib:utils:locker:withLock","time":"2020-12-02T22:33:38.920166081Z"}
```
But BB_BUILD_319 not idea which one is it...

itsjackgardner:
@RajaBellebon Yeah not sure about that one 🤔, the tests are a little flaky sometimes so that might have been a once off, looks like I tested a couple times that day so looks fine.

itsjackgardner (approved):

itsjackgardner:
@RajaBellebon I've found the error, I suspect due to some change in the BB pipelines `stopPipeline` API our handling of the failed case doesn't work anymore.
Should be an easy change to fix.
The more worrying issue is that the tests didn't catch the problem, I will look into this. For now I'll DM you with the change that has to be made :)

RajaBellebon:
Hey @itsjackgardner ,

it looks better now

Screen Shot 2020-12-07 at 14 09 10

itsjackgardner (approved):

lukebatchelor:
Sweet, nice one @RajaBellebon . Thanks.

Just wondering, what was the fix in 34a401553fa6de0bda1f5a0fc1b221d46daf36d5 ? Looks like it just added logging, or am I missing something?

RajaBellebon:
Hey @lukebatchelor ,

after talking to @itsjackgardner due to the result of the first round of integration tests, the way we handle the failed case did not work anymore. After some digging, it was due to some changes in the BB API `stopPipeline`, @itsjackgardner was telling so when a build fails due to deps, we go to stop it’s pipeline. The request was failing with 400 for some reason now, which is preventing the failed build from relanding. So i warp it up in a try and catch and return false when we got an error and log it.

PR #94, opened 2021-01-31
Afp 2658 - Adding custom data metrics

alaczko-atlassian:
Added custom data metrics for:
- on startup
- pull request landed / queued / queued when able
- user gets the current state

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -17,6 +17,7 @@ import { routes } from './routes';
 import { initializePassport } from './auth/bitbucket';
 import { LandRequestHistory } from './lib/History';
 import { Logger } from './lib/Logger';
+import { stats } from './lib/Stats';
 
 const RedisStore = connectRedis(session);
 
@@ -62,6 +63,8 @@ async function main() {
   const runner = new Runner(queue, history, client, config);
 
   await routes(server, client, runner);
+
+  stats.increment('startup');
 
   // TODO: lookup all admins in user service to add them to the redis cache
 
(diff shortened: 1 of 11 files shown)


src/lib/Stats.ts
```diff
@@ -0,0 +1,25 @@
...
+  client = {
+    increment: (stat: string, tags?: StatsD.Tags) => {},
+  };
+} else {
+  const HOST = NODE_ENV === 'production' ? 'platform-statsd' : 'localhost';
+
+  client = new StatsD.StatsD({
+    host: HOST,
+    port: 8125,
+    prefix: 'atlassian_frontend_landkid.',
```
lukebatchelor: Just double checking, is it normal to have the trailing `.` here?

src/index.ts
```diff
@@ -63,6 +64,8 @@ async function main() {
 
   await routes(server, client, runner);
 
+  stats.increment('.startup');
```
lukebatchelor, line 67: Is starting with a `.` normal too?

lukebatchelor (approved):
Sweet, lets get the PR ready to deploy to staging and make sure that's working.

src/lib/Stats.ts
alaczko-atlassian: I forgot to change that but will fix it up in the next commit with the update duration metric name.

src/index.ts
alaczko-atlassian: same as above ^ I'm planning to append the `.` to the prefix if it exists when the StatsD client is initialised, rather than have `.`s added everywhere.

PR #102, opened 2021-03-29
Move pipeline link into card title. Link pr title instead of pr number

ReDrUm:
### Move pipeline link into card title. Link PR title instead of PR number

* Moved the Pipelines Link into the card title.
  * The pipe separator in the title is deliberately light to make it fade into the background since it doesn't need to be too prominent as a separator.
* Removed the [PR #] prefix and instead hyperlinked the PR title.
  * Applied custom styling to the PR title to avoid double blue adjacent links

**Before:**

Screen Shot 2021-03-29 at 2 27 56 pm

**After:**

Screen Shot 2021-03-29 at 1 41 47 pm

### DIFF
--- a/src/static/current-state/components/QueueItem.tsx
+++ b/src/static/current-state/components/QueueItem.tsx
@@ -5,7 +5,7 @@ import { Lozenge } from './Lozenge';
 import { LozengeAppearance } from './types';
 import { User } from './User';
 
-let queueItemStyles = css({
+const queueItemStyles = css({
   display: 'block',
   boxSizing: 'border-box',
   padding: '14px 14px 14px',
@@ -96,7 +96,7 @@ let queueItemStyles = css({
   },
 });
 
-let queueItemJoinedStyles = css({
+const queueItemJoinedStyles = css({
   paddingTop: '27px',
   position: 'relative',
   '&:before': {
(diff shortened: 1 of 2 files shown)


ReDrUm:
Screen Shot 2021-03-29 at 8 11 01 pm

The pipe separator looked a bit weird on some of them due to sub-pixel rendering antialiasing blurring, so I switch it back to regular font weight and dulled the colour instead. That solved the blurring issue and I think it looks better now anyway.

itsjackgardner (approved):

src/static/current-state/components/QueueItem.tsx
```diff
@@ -320,6 +322,8 @@ export class QueueItem extends React.Component<QueueItemProps, QueueItemState> {
       request: { dependsOn, pullRequestId, pullRequest },
     } = status;
 
+    const buildId = status.request.buildId;
```
lukebatchelor, line 325: Huh... I could have sworn the reason we didn't have the build url in the title to begin with was that we didn't have the build status's info for every item in the queue, we fetched it one at a time as you clicked "show more".

Is my memory of this just wrong or did this change at some point? Do you recall @itsjackgardner ?

ReDrUm: We seem to, as far as I've observed 😅

PR #103, opened 2021-04-05
AFP-2909 Emit metrics events with metadata

luciaquirke:


Blasz (commented):
Looks good.

The events seem to be pretty basic for the most part, presumably due to statsd only allowing numeric values.
Since we're able to send richer data structures for the other metrics we want to instrument, I think it would be good to add as much metadata as possible to each event emitted. For example: branch name, target branch, commit etc.
We could then use this metadata to link it with other metadata we collect for branches in CI, such as number of changed packages etc.

If we went down this path, I'm wondering whether we'd want to strictly type the different event types and their payloads or not. It looks like landkid configuration files are javascript. Thoughts @itsjackgardner?

src/types.ts
```diff
@@ -65,10 +65,17 @@ export type OAuthConfig = {
...
-  host: string;
-  port: number;
-  prefix?: string;
+export type EventData = {
+  landRequestId?: string;
+  pullRequestId?: string;
+  targetBranch?: string;
+  commit?: string;
+  duration?: number;
+};
```
Blasz, line 74: Nice, maybe `sourceBranch` would be worthwhile also in addition to `targetBranch`

Blasz (approved):
LGTM. Lets wait for @itsjackgardner's review as he's more familiar with the codebase.

src/lib/Events.ts
```diff
@@ -0,0 +1,18 @@
...
+import { config } from '../lib/Config';
+import { Logger } from '../lib/Logger';
+import { EventData } from '../types';
+
+export const eventEmitter = new EventEmitter();
+
+export const initializeEventListeners = () => {
+  if (config.eventListeners) {
+    config.eventListeners.forEach(({ event, listener }) => {
+      eventEmitter.addListener(event, (data: EventData) => {
```
itsjackgardner, line 12: Could we change the function that's passed into the `.addListener` to be:

```ts
(data?: EventData) => {
  Logger.info(`Emitting event ${event}`, { data });
  listener(data || {});
}
```

That way we can emit events without having to include metadata if there isn't any (e.g. `eventEmitter.emit('STARTUP')`), and it just defaults to an empty object.

src/lib/Runner.ts
```diff
@@ -556,17 +567,6 @@ export class Runner {
...
-  getLandRequestQueuedDate = async (requestId: string): Promise<Date | null> => {
-    const status = await LandRequestStatus.findOne<LandRequestStatus>({
-      where: {
-        requestId,
-        state: 'queued',
-      },
-      order: [['date', 'ASC']],
-    });
-    return status ? status.date : null;
-  };
```
itsjackgardner, line 568: I just realised that I was wrong about not needing this function anymore, will need to replace `landRequestStatus.date.getTime()` with what we had before, sorry about that.

itsjackgardner (changes requested):

PR #109, opened 2021-04-26
Add `skipBuildOnDependentsAwaitingMerge` config option

Blasz:
When set, merge commits for PRs will opt out of the destination branch
build via [skip ci] when there are successful dependent PRs awaiting
merge.
This prevents multiple destination branch builds from triggering when
PRs are landed in quick succession which can affect builds that update
src and push back to the repo such as publishing new versions of npm
libraries.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -94,6 +94,16 @@ export type Config = {
   sequelize?: any;
   eventListeners?: EventListener[];
   easterEgg?: any;
+  mergeSettings?: MergeSettings;
+};
+
+export type MergeSettings = {
+  /** Skip the destination branch build when there are successful dependent requests awaiting merge.
+   * This prevents multiple branch builds triggering multiple merges happen in quick succession.
+   * Achieved by adding [skip ci] to the merge commit message
+   */
+  skipBuildOnDependentsAwaitingMerge?: boolean;
+  // waitForBuild?: { // TBD };
 };
 
 export type RunnerState = {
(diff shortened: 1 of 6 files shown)


src/types.ts
```diff
@@ -106,3 +116,7 @@ export type RunnerState = {
   bitbucketBaseUrl: string;
   permissionsMessage: string;
 };
+
+export type MergeOptions = {
```
luciaquirke, line 120: Where does this type get used?

itsjackgardner: Here: https://github.com/atlassian/landkid/pull/109/files#diff-31195c1d7289295f28e7902fd02c31cad4122633ef305054421ae836ce368f63R17

src/types.ts
```diff
@@ -94,6 +94,16 @@ export type Config = {
   sequelize?: any;
   eventListeners?: EventListener[];
   easterEgg?: any;
+  mergeSettings?: MergeSettings;
+};
+
+export type MergeSettings = {
+  /** Skip the destination branch build when there are successful dependent requests awaiting merge.
+   * This prevents multiple branch builds triggering multiple merges happen in quick succession.
+   * Achieved by adding [skip ci] to the merge commit message
```
luciaquirke, line 103: What consumes the [skip ci] commit message?

itsjackgardner: Hmm that is interesting, but not a concern here since it affects only the build that runs on the target branch after merging the PR.

itsjackgardner (approved):

src/types.ts
luciaquirke: LGTM :)

luciaquirke (approved):

PR #110, opened 2021-04-27
New state for when a PR is in the process of merging

itsjackgardner:
This avoids the issue where merging takes longer than the TTL of the `status-transition` lock which allows another merge attempt to be triggered, causing the land request to say "Unable to merge pull request" despite having successfully merged it.

### DIFF
--- a/src/lib/Runner.ts
+++ b/src/lib/Runner.ts
@@ -229,26 +229,34 @@ export class Runner {
         landRequestId: landRequest.id,
         lockId,
       });
-      await this.client.mergePullRequest(landRequestStatus, { skipCI });
-      Logger.info('Successfully merged PR', {
+
+      await landRequest.setStatus('merging');
+      this.client.mergePullRequest(
+        landRequestStatus,
+        async () => {
+          const end = Date.now();
+          const queuedDate = await this.getLandRequestQueuedDate(landRequest.id);
+          const start = queuedDate!.getTime();
+          eventEmitter.emit('PULL_REQUEST.MERGE.SUCCESS', {
+            landRequestId: landRequestStatus.requestId,
+            pullRequestId: landRequest.pullRequestId,
+            commit: landRequest.forCommit,
+            sourceBranch: pullRequest.sourceBranch,
+            targetBranch: pullRequest.targetBranch,
+            duration: end - start,
+          });
+          landRequest.setStatus('success');
+        },
+        { skipCI },
+      );
+
+      Logger.info('Moved request to merging', {
         namespace: 'lib:runner:moveFromAwaitingMerge',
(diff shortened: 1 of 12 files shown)


itsjackgardner:
Screen Shot 2021-04-29 at 11 56 03 am

src/lib/Runner.ts
```diff
@@ -223,32 +223,34 @@ export class Runner {
...
+            duration: end - start,
+          });
+          landRequest.setStatus('success');
+        },
+        { skipCI },
+      );
+
+      return true;
     } catch (err) {
       eventEmitter.emit('PULL_REQUEST.MERGE.FAIL', {
```
Blasz, line 255: Will this `catch` block catch errors thrown in the `mergePullRequest` function and/or the callback? We might have to add a `.catch()` directly to the function call.

src/lib/Runner.ts
```diff
@@ -223,32 +223,34 @@ export class Runner {
...
-        commit: landRequest.forCommit,
-        sourceBranch: pullRequest.sourceBranch,
-        targetBranch: pullRequest.targetBranch,
-        duration: end - start,
-      });
-      return landRequest.setStatus('success');
+      await landRequest.setStatus('merging');
+      this.client.mergePullRequest(
+        landRequestStatus,
+        async () => {
```
Blasz, line 236: Is there any difference between using a callback here vs using `Promise.then`? Might be simpler to use `Promise.then`.

src/lib/utils/locker.ts
```diff
@@ -15,7 +19,7 @@ export const withLock = async <T>(resource: string, fn: (lockId: Date) => Promis
       lockId,
     });
   } catch {
-    return;
+    return fallback;
```
Blasz, line 22: What does adding the return type here do? Is there a noticeable difference between `false` and `undefined` in our `withLock` usages above?

Blasz (commented):
Nice one, looks good. Only major concern is whether we properly catch merge failures with the new callback logic.

src/lib/Runner.ts
itsjackgardner: Yeah probably cleaner to rework the `try...catch` into using `.then().catch()` so we can also make sure that errors are caught.

src/lib/utils/locker.ts
itsjackgardner: I added it when I was making a couple other changes to the lock that I reverted, but I think its cleaner with better defined return values.

itsjackgardner:
Screen Shot 2021-04-29 at 2 55 01 pm

Tested again with succeed and fail states.

Blasz (approved):

PR #113, opened 2021-05-04
Actually use merging state properly

itsjackgardner:


RajaBellebon (approved):
LGTM, is there any unit tests that would cover this?

src/lib/Queue.ts
```diff
@@ -1,5 +1,16 @@
 import { LandRequestStatus, LandRequest, PullRequest } from '../db';
 
+const order = [
```
RajaBellebon, line 3: where are you using order?

src/lib/Queue.ts
```diff
@@ -17,14 +28,14 @@ export class LandRequestQueue {
...
-  // returns the list of queued, running and awaiting-merge items as these are the actual "queue" per se
+  // returns the list of queued, running, awaiting-merge, and merging items as these are the actual "queue" per se
   // all the status' we display on the frontend
   public getQueue = async (): Promise<LandRequestStatus[]> => {
     const queue = await LandRequestStatus.findAll<LandRequestStatus>({
       where: {
         isLatest: true,
         state: {
-          $in: ['queued', 'running', 'awaiting-merge'],
+          $in: ['queued', 'running', 'awaiting-merge', 'merging'],
```
RajaBellebon, line 38: do we need `'will-queue-when-ready',` too?

itsjackgardner: Nope this is just what should appear on the `current-state` screen.

src/lib/Queue.ts
itsjackgardner: Here: https://github.com/atlassian/landkid/pull/113/files#diff-2104577a08320b88076fc9c0cb711a4a6de3ce58e10bfa067afb271d21db5149R50

itsjackgardner:
> LGTM, is there any unit tests that would cover this?

@RajaBellebon  Potentially, will need to look into it later though.

PR #118, opened 2021-05-19
New merge poller

itsjackgardner:
New class that manages merging PRs, as well as polling the merge response.


tests/integration/double-reentry.spec.ts
```diff
@@ -19,9 +19,9 @@ describe('Re-entry into queue', () => {
   });
 
   it('Requests are re-entered into queue after the failure of dependency', async () => {
-    assert(failed.validate(prs[branch1].statuses));
-    assert(reentryFail.validate(prs[branch2].statuses));
-    assert(doubleReentrySuccess.validate(prs[branch3].statuses));
+    assert(await validate(prs[branch1].statuses, failed));
+    assert(await validate(prs[branch2].statuses, reentryFail));
+    assert(await validate(prs[branch3].statuses, doubleReentrySuccess));
```
itsjackgardner, line 24: Turns out the statuses weren't actually. being verified properly, but it usually didn't matter because these tests just existed to ensure nothing broke on the dev instance of Landkid so a quick manual verification was all that was required.

tests/unit/BitbucketAPI.test.ts
```diff
@@ -85,10 +68,13 @@ describe('mergePullRequest', () => {
...
       .mockResolvedValueOnce({ data: { task_status: 'PENDING' } })
       .mockRejectedValueOnce({ response: {} });
-    await expect(mergePullRequest(landRequestStatus)).rejects.toThrow();
-    expect(loggerInfoSpy).toHaveBeenCalledTimes(2);
-    expect(loggerErrorSpy).toHaveBeenCalledTimes(1);
-    expect(mockedDelay).toHaveBeenCalledTimes(2);
+    expect(await mergePullRequest(landRequestStatus)).toEqual(BitbucketAPI.FAILED);
+  });
+
+  test('Timeout merge polling', async () => {
```
RajaBellebon, line 74: are you testing the counter as well?

RajaBellebon (approved):

tests/unit/BitbucketAPI.test.ts
itsjackgardner: This is testing the max attempts, but I couldn't figure out how to test the abort thing, because it's a function that's called while the async merging is occurring.

Blasz (approved):
LGTM.

One existing case I'm not sure that we handle is the `merging` state if the landkid instance is redeployed or restarts for whatever reason. It looks like it could get stuck in that state since the polling is all done in memory. Of course the new ability to manually abort will allow us to resolve this but I'm wondering whether we should also try to resolve this case automatically as well.

tests/unit/BitbucketAPI.test.ts
Blasz: You might be able to do it by using normal promises rather than async/await.

```
const mergePromise = mergePullRequest(landRequestStatus).then(res => { expect(res).toEqual(BitbucketAPI.ABORT); });
bitbucketAPI.cancelMergePolling(...);
return mergePromise;
```

Could also return the `done` callback parameter instead of returning the promise as well - https://jestjs.io/docs/asynchronous

itsjackgardner: Yeah I attempted this and it didn't work. Will merge for now and potentially come back to that later.

PR #119, opened 2021-05-21
Land request priority system

itsjackgardner:
Just a simple numeric priority system, stored in the request table so it can persist between queue statuses if something re-queues due to a failed dependency.

Admins can increment and decrement based on the importance of a PR, and requests will be bumped every time they re-queue so that merge times are decreased for PRs unfairly effected by other PRs.

### DIFF
--- a/src/routes/api/index.ts
+++ b/src/routes/api/index.ts
@@ -8,6 +8,7 @@ import { AccountService } from '../../lib/AccountService';
 import { permissionService } from '../../lib/PermissionService';
 import { Config, LandRequestOptions } from '../../types';
 import { eventEmitter } from '../../lib/Events';
+import { LandRequest } from '../../db';
 
 const landKidTag = process.env.LANDKID_TAG || 'Unknown';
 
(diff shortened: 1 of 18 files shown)


src/routes/api/index.ts
```diff
@@ -303,19 +313,24 @@ export function apiRoutes(runner: Runner, client: BitbucketClient, config: Confi
...
+        return res.status(400).json({ err: 'req.body.priority expected', body: req.body });
+      }
+      const { priority } = req.body;
+      Logger.verbose('Updating priority of land request', {
+        namespace: 'routes:api:priority',
         requestId,
+        priority,
       });
-      const success = await runner.moveRequestToTopOfQueue(requestId, req.user!);
+      const success = await runner.updateLandRequestPriority(requestId, priority);
```
Blasz, line 329: What happens if priority is an invalid type, does it just 500?

Blasz (approved):
Nice one, LGTM. Is there a staging link where we can see this in action?

src/routes/api/index.ts
itsjackgardner: Hmm, it will 500 but would probably be better if it responded with 400 instead.

itsjackgardner:
> Nice one, LGTM. Is there a staging link where we can see this in action?

It's live here: https://atlassian-frontend-landkid.dev.services.atlassian.com/current-state/

PR #121, opened 2021-05-31
Remove unnecessary increment priority button

itsjackgardner:
Don't know how I didn't notice this but the button I was using to test the API was left in after I added in the real buttons.


luciaquirke (approved):
LGTM

PR #124, opened 2021-07-01
Allow starting on different ports

lukebatchelor:
This PR adds `SERVER_PORT` and `DEV_SERVER_PORT` variables to allow starting the server on different ports

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -4,6 +4,8 @@ const HtmlWebpackPlugin = require('html-webpack-plugin');
 
 // This only really matters when building files, not in production
 const outputPath = process.env.OUTPUT_PATH || '.';
+const SERVER_PORT = process.env.SERVER_PORT || '8080';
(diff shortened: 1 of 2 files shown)


webpack.config.js
```diff
@@ -20,14 +22,14 @@ module.exports = {
     compress: true,
     historyApiFallback: true,
     // hot: true,
-    port: 3000,
+    port: Number(PROXY_PORT),
```
ReDrUm, line 25: Based on our conversation yesterday, perhaps we should rename this to `DEV_SERVER_PORT`.

> "I still don't love "PROXY_PORT". It should really be something like DEV_SERVER_PORT because it's only used in dev. The other port is used internally during dev and externally during production (in production the node server serves the static assets, in development the frontend server proxies requests to the backend. It's v. weird)" - @lukebatchelor

ReDrUm (approved):

luciaquirke (approved):

PR #127, opened 2021-07-26
Display more verbose error with instructions when user denies access to the addon

itsjackgardner:
Screen Shot 2021-07-26 at 3 20 48 pm

### DIFF
--- a/src/bitbucket/descriptor.ts
+++ b/src/bitbucket/descriptor.ts
@@ -1,80 +1,93 @@
 import { config } from '../lib/Config';
 
+const getAppName = () =>
+  `${config.name || 'Landkid'} ${(process.env.LANDKID_DEPLOYMENT || 'local').toUpperCase()}`;
+
 export const getAppKey = () =>
   `${config.key || 'landkid'}-${process.env.LANDKID_DEPLOYMENT || 'local'}`;
 
-export const makeDescriptor = () => ({
-  name: `${config.name || 'Landkid'} ${(process.env.LANDKID_DEPLOYMENT || 'local').toUpperCase()}`,
-  description: "Addon to display a 'release queue' panel in PRs",
-  baseUrl: config.baseUrl,
-  key: getAppKey(),
-  vendor: {
-    name: 'Fabric Build',
-  },
-  // :write implies base scope (e.g. pullrequest:write implies pullrequest)
-  // pullrequest:write requires repository:write so we don't need to define that as well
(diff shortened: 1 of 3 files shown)


src/bitbucket/descriptor.ts
```diff
@@ -1,80 +1,93 @@
...
+
+  const addonName = `Landkid Queue${
+    process.env.LANDKID_DEPLOYMENT !== 'prod'
+      ? ` (${process.env.LANDKID_DEPLOYMENT || 'local'})`
+      : ''
+  }`;
+
+  const params = new URLSearchParams();
+  params.append('appName', appName);
+  const appNameQueryString = params.toString();
```
itsjackgardner, line 21: Passing the app name in as a query string param was the only way I could figure out to access it from the addon.

src/static/bitbucket/components/App.tsx
```diff
@@ -22,18 +31,22 @@ type AppState = {
   bannerMessage: BannerMessage | null;
 };
 
-export class App extends React.Component {
+const qs = new URLSearchParams(window.location.search);
+
+export class App extends React.Component<{}, AppState> {
+  appName = qs.get('appName') || 'Landkid';
```
itsjackgardner, line 37: Maybe there's a better default (i.e. point to the app name displayed above the addon)?

src/bitbucket/descriptor.ts
lukebatchelor: Hacky solution, but could we bake it into `src/static/bitbucket/components/App.tsx` directly? Would need to do something like adding it as an env var at build time though. Which is super doable, but is it more or less annoying that getting the addon refreshed? 🤔

src/bitbucket/descriptor.ts
```diff
@@ -1,80 +1,93 @@
...
+    // pullrequest:write requires repository:write so we don't need to define that as well
+    scopes: ['account', 'pullrequest:write', 'pipeline:write'],
+    authentication: {
+      type: 'jwt',
     },
-    webPanels: [
-      {
-        weight: 100,
-        tooltip: {
-          value: 'Packages to be released in this PR',
```
lukebatchelor, line 43: Ha!

src/static/bitbucket/components/App.tsx
lukebatchelor: I think this is fine tbh.

lukebatchelor:
Sweet, lgtm!

lukebatchelor (approved):

PR #134, opened 2021-09-14
Revert "Update docker image version"

luciaquirke:
Reverts atlassian/landkid#132


luciaquirke:
Original pull request appeared to be merging in content already present in master, but after merging it and inspecting its complementary revert PR, this no longer appears to be the case. Putting it down to a big in Github

PR #136, opened 2021-09-23
add workaround for getting task count

diffidentDude:
This change introduces a workaround for an issue where bitbucket returns a task_count of 0 when calling the pull_request API. We can instead directly call the /tasks endpoint, which returns the correct data.

This fix will unblock https://hello.atlassian.net/browse/EXPAND-1437
The bitbucket bug is tracked https://softwareteams.atlassian.net/browse/COREX-4320


RajaBellebon (commented):
Changes look good, just one question why do we not compute the number of task and store it into a variable outside of the return object

PR #140, opened 2021-11-15
WIP AFP-3330 only fetch requests from last 7 days

cheesehary:
only fetch requests from last 7 days

### DIFF
--- a/src/lib/Queue.ts
+++ b/src/lib/Queue.ts
@@ -1,4 +1,5 @@
 import { LandRequestStatus, LandRequest, PullRequest } from '../db';
+import { Op } from 'sequelize';
 
 export class LandRequestQueue {
   public getStatusesForWaitingRequests = async () => {
@@ -6,6 +7,8 @@ export class LandRequestQueue {
       where: {
         isLatest: true,
         state: 'will-queue-when-ready',
+        // only retrieve requests from last 7 days
+        date: { [Op.gt]: new Date(new Date().getTime() - 7 * 1000 * 60 * 60 * 24) },
       },
       order: [['date', 'ASC']],
       include: [


src/lib/Queue.ts
```diff
@@ -1,11 +1,14 @@
...
+import { Op } from 'sequelize';
 
 export class LandRequestQueue {
   public getStatusesForWaitingRequests = async () => {
     return LandRequestStatus.findAll<LandRequestStatus>({
       where: {
         isLatest: true,
         state: 'will-queue-when-ready',
+        // only retrieve requests from last 7 days
+        date: { [Op.gt]: new Date(new Date().getTime() - 7 * 1000 * 60 * 60 * 24) },
```
RajaBellebon, line 11: Just curious, how did you find out that was the root cause of the issue?

cheesehary: In the [log](https://splunk.paas-inf.net/en-US/app/search/search?q=search%20m.t%3Dapplication%20%60micros_atlassian-frontend-landkid%60%20env%3Dprod-west%20NOT%20message%3D*JWT*%20(namespace%3Dlib%3A*%20OR%20namespace%3Dbitbucket%3Aclient%3A*)%20(pullRequestId%3D16522%20OR%20NOT%20pullRequestId%3D*)&earliest=1636763760.926&latest=1636763972.956&display.page.search.mode=verbose&dispatch.sample_ratio=1&sid=1636944668.1718420_62D9EEBE-1751-48A5-B0C6-CDABAC40F831&display.prefs.events.offset=20), we found that the **waitingRequestStatuses** shows there are over 100 requests with state being **will-queue-when-ready** and date of the oldest one is Sep 2020. 

These outdated requests get fetched every time and very likely cause a race condition. Because **checkWaitingLandRequests** can't finish all requests in the queue in 2 minutes, in the second run of **checkWaitingLandRequests**, our latest request will trigger the abort status.

This PR only handles half of the issues by not fetching outdated requests. We'll need to address the race condition issue as well. But I'm not sure how to test the change in dev/local environment, I followed the [guide](https://bitbucket.org/atlassian/atlassian-frontend-landkid-deployment/src/master/development.md) but am currently blocked

RajaBellebon (approved):

src/lib/Queue.ts
RajaBellebon: @cheesehary do you know if we could add a unit test for this use case or is that too complex?

cheesehary:
Manual Testing:

1. set the time from 7 days to 1 hour for faster testing
2. create a land request with state **will-queue-when-ready**, decline the attached PR so that it will stay **will-queue-when-ready**
3. check the **waitingToQueue** of the response from **/api/current-state**, should find the newly created land request 
4. after 1 hour, check the response again, the land request should disappear

src/lib/Queue.ts
cheesehary: a unit test probably requires us to mock most of the logic and we'll end up proving a smaller queue takes less time, which might not worth it.

PR #144, opened 2022-02-10
Give authenticated users land rights by default

luciaquirke:
This needs to be combined with an update to the Landkid deployment repositories, so Landkid is deployed behind the VPN.

### DIFF
--- a/src/db/index.ts
+++ b/src/db/index.ts
@@ -241,7 +241,7 @@ export class Permission extends Model<Permission> implements IPermission {
   readonly aaid: string;
 
   @AllowNull(false)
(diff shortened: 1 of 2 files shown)


RajaBellebon:
Do we need to modify this one too https://github.com/atlassian/landkid/blob/master/src/db/index.ts#L244

RajaBellebon (changes requested):
Can we check tif we need to modify this DB too https://github.com/atlassian/landkid/blob/master/src/db/index.ts#L244

RajaBellebon (approved):

PR #149, opened 2022-03-14
Slack bot innovation week

Blasz:
This PR makes a few changes to support sending slack notifications when the landkid status of your PR changes.
The main changes are:

* Emitting a new `'LAND_REQUEST.STATUS.CHANGED'` event when a land request status is updated. We can then add an event listener in our own instance for it that then sends a request to a new `af-notifications` service (will create a PR for this later) that is responsible for sending the slack notification.
* Sending & storing atlassian account IDs of both PR authors and land request triggerers. The current `aaid` fields that we send & store were actually changed to bitbucket UUIDs a very long time ago but were never renamed to reflect this.

Some things to call out:

* The real aaid fields are labelled `accountId` to differentiate them from the existing misnamed aaid fields. Renaming the fields would be a migration headache.
* For the same migration headache reasons, the new fields have been configured to allow null values. This means we don't have to update existing entries in our databases with non-null values and follows the pattern of previous migrations we've added.
* We need to re-install the landkid bitbucket connect add-on as the service descriptor has been updated to send the actual aaid of the triggerer in the land requests.


src/db/index.ts
```diff
@@ -108,6 +124,20 @@ export class LandRequest extends Model<LandRequest> implements ILandRequest {
         { transaction: t },
       );
     });
+    eventEmitter.emit('LAND_REQUEST.STATUS.CHANGED', {
+      landRequestId: this.id,
+      pullRequestId: this.pullRequestId,
+      buildId: this.buildId,
+      author: this.pullRequest?.authorAccountId,
```
RajaBellebon, line 131: is there a difference between account_id and authorAccountId

src/lib/Runner.ts
```diff
@@ -187,11 +187,11 @@ export class Runner {
         dependencies.map((queueItem) => queueItem.request.pullRequestId).join(', ');
     }
 
-    await landRequest.setStatus('running', depPrsStr);
-
     // Todo: these should really be functions on landRequest
     landRequest.buildId = buildId;
     landRequest.dependsOn = dependsOnStr;
+    await landRequest.setStatus('running', depPrsStr);
```
RajaBellebon, line 193: any reason for moving the function?

src/lib/Runner.ts
```diff
@@ -490,11 +491,17 @@ export class Runner {
...
-    });
+    return LandRequest.create<LandRequest>(
+      {
+        triggererAaid: landRequestOptions.triggererAaid,
+        triggererAccountId: landRequestOptions.triggererAccountId,
+        pullRequestId: pr.prId,
+        forCommit: landRequestOptions.commit,
+      },
+      {
+        include: [PullRequest],
```
RajaBellebon, line 502: are we adding more metadata?

src/routes/api/index.ts
```diff
@@ -261,9 +261,13 @@ export function apiRoutes(runner: Runner, client: BitbucketClient, config: Confi
         prId,
         // Must always have a valid aaid, so lets set it to Luke's aaid
         triggererAaid: req.user ? req.user.aaid : '557057:9512f4e4-3319-4d30-a78d-7d5f8ed243ae',
+        triggererAccountId: req.user
+          ? req.user.accountId
+          : '557057:9512f4e4-3319-4d30-a78d-7d5f8ed243ae',
         commit: prInfo.commit,
         prTitle: prInfo.title,
         prAuthorAaid: prInfo.authorAaid,
+        prAuthorAccountId: prInfo.author,
```
RajaBellebon, line 270: any reason to not be `authorAccountId`

RajaBellebon (approved):
The changes lgtm - left few minor questions

luciaquirke (approved):

PR #150, opened 2022-03-23
when paused, disallow land requests to move to running

cheesehary:
land requests won't move from queued to running when paused

### DIFF
--- a/src/lib/Runner.ts
+++ b/src/lib/Runner.ts
@@ -68,6 +68,7 @@ export class Runner {
   };
 
   moveFromQueueToRunning = async (landRequestStatus: LandRequestStatus, lockId: Date) => {
+    if (await this.getPauseState()) return;
     const landRequest = landRequestStatus.request;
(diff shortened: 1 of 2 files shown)


src/lib/Runner.ts
```diff
@@ -68,6 +68,7 @@ export class Runner {
   };
 
   moveFromQueueToRunning = async (landRequestStatus: LandRequestStatus, lockId: Date) => {
+    if (await this.getPauseState()) return;
```
RajaBellebon, line 71: should we hold this value in a variable instead?

RajaBellebon (approved):

src/lib/Runner.ts
cheesehary: We still need to put the variable in redis, which might be a bit quicker than db. But the `PauseState` is an one-row table, I guess the query is not that slow and we won't get a drastic speed increase by moving to redis.

PR #153, opened 2022-05-19
ShipIt 54: Add-on theming updates

luke-underwood:
## These changes are part of a ShipIt 54 project SHPLIV-169.
**Theming adjustments to add-on to align styles closer with Atlassian Design System guidelines. This makes the add-on look like it belongs in Bitbucket.**

- Replaced UI with Atlassian Design System components
- Installed Storybook to improve add-on development experience. [Demo video here](https://www.loom.com/share/8a0b8ce13c2d4c77938a81061ab64405). Start using `yarn run storybook`
- Splits class component App.tsx into several smaller functional components. The separation makes it much easier to test in isolation, as Bitbucket is not required to develop the UI of the add-on
- Added loading states when performing actions "land" and "land when able". This provides feedback to the user that something is happening once they've clicked a button
- Added loading spinner while checking if able to land
- Fixed height to 200px with a scrollbar for overflow. This prevents issues caused by slow load times. Unfortunately Landkid's `/can-land` is usually the last thing to load on pull requests with times around 15 seconds. Transitioning from the loading state altered the height of the add-on, pushing the rest of the pull request downwards which causes frustrating cumulative layout shift. The fixed height prevents this happening. This is not an ideal solution but I think it's necessary unless we can improve response times from the API
- Error wording updates

### Before
landkid-before

### After
landkid-light-mode

A bonus of using Atlassian Design System components: It's Dark mode ready (won't be usable until supported in Bitbucket)

landkid-dark-mode

PR #166, opened 2022-08-31
ShipIt 54: Build fixes

luke-underwood:
Following https://github.com/atlassian/landkid/pull/153, there were some TypeScript and Webpack issues, mostly due to adding Storybook.

I bumped a few versions to resolve. Builds are working again, but if there's too many dependency changes here let me know and I'll consider ripping Storybook back out instead 👍 

- Fixed TSconfig issues
- Prevent Storybook stories emitting in `/dist`
- Increased fixed height of plugin (before it vertically scrolls) to prevent text cut off
- Added `style-loader` and `css-loader` for `@atlaskit/css-reset`
- Upgraded to Webpack 5, as well as upgrading loaders
- Bumped `react` to `^16.8.0`
- Added confetti when you click **Land** 🥳🎉🎊


Brianzchen (approved):

pastelsky:
The confetti would only be rendered within the iframe bounds, so you might have some cropping going on.

luke-underwood:
@pastelsky It's unfortunate but I think it looks okay regardless. Maybe it's good the confetti can't interfere with anything else on the page 🤔 

I tried making it smaller and more within the boundaries, but it didn't have the same visual effect

PR #167, opened 2022-08-31
Parallelize can land checks

pastelsky:
Parallelize API calls and checks that happen in the `can-land` pipeline. Should speed up the `can-land` endpoint by about 2x for our monorepo checks.


package.json
```diff
@@ -78,7 +78,7 @@
     "prettier": "^2.3.0",
     "rimraf": "^2.6.2",
     "rxjs-compat": "^6.3.3",
-    "sqlite3": "^5.0.0",
+    "sqlite3": "^5.0.11",
```
pastelsky, line 81: Bumping this up to add compatibility for Mac M1

Brianzchen (approved):

Brianzchen:
We should probably add more tests to this repo.. 🤔

pastelsky:
Yeah, that'd been ideal. More than the tests though,  I think having a reasonable devloop for testing changes like these would be good since testing this is a massive pain at the moment.

PR #169, opened 2022-08-31
Tweak formatting of landkid message

pastelsky:
Tweaks section message formatting —
1. Use shimmer as placeholder instead of setting a fixed height —
Screenshot 2022-08-31 at 9 11 55 PM

2. Moves errors and warnings inside the main section message —
(They looked a little odd outside, given that bitbucket doesn't add separators between the landkid widget and the next one)
Screenshot 2022-08-31 at 9 12 41 PM

3. Makes the land button bigger, moves it into main section message —
(same reason as 2.)
Screenshot 2022-08-31 at 9 13 30 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -90,7 +90,6 @@
   "dependencies": {
     "@atlaskit/button": "^15.0.0",
     "@atlaskit/css-reset": "^6.3.13",
-    "@atlaskit/menu": "^1.3.9",
     "@atlaskit/pagination": "^8.0.5",
     "@atlaskit/section-message": "^6.1.12",
     "@atlaskit/spinner": "^15.1.11",
--- a/package.json
+++ b/package.json
@@ -90,6 +90,7 @@
   "dependencies": {
     "@atlaskit/button": "^15.0.0",
     "@atlaskit/css-reset": "^6.3.13",
+    "@atlaskit/menu": "^1.3.9",
     "@atlaskit/pagination": "^8.0.5",
     "@atlaskit/section-message": "^6.1.12",
     "@atlaskit/spinner": "^15.1.11",
--- a/src/static/bitbucket/components/Message.tsx
+++ b/src/static/bitbucket/components/Message.tsx
@@ -5,10 +5,12 @@ import SectionMessage, {
 } from '@atlaskit/section-message';
 import { LoadingButton as Button } from '@atlaskit/button';
 import Spinner from '@atlaskit/spinner';
+import { SkeletonItem } from '@atlaskit/menu';
 import Confetti from 'react-dom-confetti';
 
 import Errors from './Errors';
 import Warnings from './Warnings';
+import type { ReactElement } from 'react';
 
 type Status =
   | 'checking-can-land'
@@ -32,7 +34,7 @@ const messageAppearance: { [keyof in Status]: SectionMessageProps['appearance'] 
 } as const;
 
 const messageTitle: { [keyof in Status]: string } = {
-  'checking-can-land': 'Loading Landkid...',
+  'checking-can-land': 'Checking land status...',
   'cannot-land': 'Not ready to land',
   queued: 'Queued to land!',
   'can-land': 'Ready to land!',
@@ -87,6 +89,13 @@ const Message = ({
   const renderLandState = () => {
     switch (status) {
       case 'checking-can-land': {
+        return (
+          <>
+            <SkeletonItem isShimmering />
+            <SkeletonItem isShimmering />
+            <SkeletonItem isShimmering />
+          </>
+        );
         return <Spinner />;
       }
       case 'cannot-land': {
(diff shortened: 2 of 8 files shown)


src/static/bitbucket/components/Warnings.tsx
```diff
@@ -1,14 +1,24 @@
+import listStyles from './styles/listStyles';
+
 type WarningProps = {
   warnings: string[];
 };
 
+import styled from 'styled-components';
```
Brianzchen, line 7: why are we mixing styled-components?

package.json
```diff
@@ -90,6 +90,7 @@
   "dependencies": {
     "@atlaskit/button": "^16.3.6",
     "@atlaskit/css-reset": "^6.3.13",
+    "@atlaskit/menu": "^1.3.9",
```
luke-underwood, line 93: I'm not sure importing this solely for the loading skeleton is best, especially since the skeleton is misaligned (there's extra padding on the left).

I think if we go down the route of loading skeletons we need to build our own as there's no existing component in the design system. There's a good one we can copy from `@atlaskit/help` called `LoadingRectangle`

luke-underwood (changes requested):

src/static/bitbucket/components/Warnings.tsx
pastelsky: Ah right, I missed converting this one to emotion. Will do.

src/static/bitbucket/components/Message.tsx
```diff
@@ -161,33 +192,9 @@ const Message = ({
         ]}
       >
         {renderLandState()}
+        {showErrors && <Errors errors={errors} />}
```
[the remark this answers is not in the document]
pastelsky: For the user to be able to act on "Land" / "Land when able" etc, reading these errors (and maybe warnings ?) might be important pieces of information. Having them outside changes the flow of information, which I thought was important. 

@Brianzchen do you think warnings are something we want users to still read and consider before they act on Land?

package.json
pastelsky: Yeah, I did feel a bit icky about this one. Will use the one you suggested.

luke-underwood: Nice! If you need any help let me know 👍

src/static/bitbucket/components/Message.tsx
```diff
@@ -115,14 +122,33 @@ const Message = ({
...
+        Land changes
+      </Button>
+    </div>
+  );
+
   return (
-    <div
-      style={{
-        // Fixed height prevents cumulative layout shift caused by slow loads
-        height: 270,
```
luke-underwood, line 122: The fixed height prevents layout shift, as Landkid is usually the last thing to load and it pushes all other content down. The loading skeleton will help but still prone to cause jumps. I feel we shouldn't remove this unless we can drastically reduce Landkid's loading times to a point where it isn't as noticeable 🤔

pastelsky: @luke-underwood The shimmer hopefully reduces the CLS to a level where it isn't a big concern.
Having a fixed height introduces a few problems —

#### White gap
Screenshot 2022-09-01 at 9 01 14 AM

#### Overlap (exaggerated by moving content inside section message)
Screenshot 2022-09-01 at 9 01 44 AM

pastelsky: >  I feel we shouldn't remove this unless we can drastically reduce Landkid's loading times to a point where it isn't as noticeable

Hopefully, the landkid status check times should be cut down by more than half after https://github.com/atlassian/landkid/pull/167 is in. Would still be noticeable though.

pastelsky: Also, added a fixed height here —
[src/static/bitbucket/index.html](https://github.com/atlassian/landkid/pull/169/files#diff-42716c2f50292cfbd69d7319e1865e2164bddd9c947c3aa7bc98f1ab0d9e6bba)

which was causing the initial CLS

Brianzchen (approved):

package.json
luke-underwood: That looks great now! Nice work.

I just rebased on master + tweaked the last line to be a bit narrower. Hope that's ok!

skeleton

src/static/bitbucket/components/Message.tsx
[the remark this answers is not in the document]
luke-underwood: Just checked it out in Storybook. I agree, warnings before the land button probably best :)

src/static/bitbucket/components/Message.tsx
luke-underwood: > Hopefully, the landkid status check times should be cut down by more than half after https://github.com/atlassian/landkid/pull/167 is in

Wow that'd be great to see. In that case I'm happy to eat the layout shift for now 👍

luke-underwood (approved):

PR #181, opened 2022-09-26
Cut down on auto-refresh

pastelsky:
Don't auto refresh when not in view — previously we'd auto-refresh if the widget was not in the viewport, albeit at a much slower rate.

### DIFF
--- a/src/static/bitbucket/components/App.tsx
+++ b/src/static/bitbucket/components/App.tsx
@@ -71,16 +71,15 @@ const App = () => {
   let refreshTimeoutId: Timeout;
 
   const pollAbleToLand = () => {
-    const isTabInForeground = !document.hidden;
-    let refreshIntervalMs = inViewRef.current ? 10000 : 20000;
+    const isInForeground = !document.hidden && inViewRef.current;
 
-    const checkPromise = isTabInForeground ? checkIfAbleToLand() : Promise.resolve();
+    const checkPromise = isInForeground ? checkIfAbleToLand() : Promise.resolve();
 
     checkPromise.finally(async () => {
       if (statusRef.current == 'pr-closed') return;
       refreshTimeoutId = setTimeout(() => {
         pollAbleToLand();
-      }, refreshIntervalMs);
+      }, 10000);
     });
   };
 
--- a/src/static/bitbucket/components/App.tsx
+++ b/src/static/bitbucket/components/App.tsx
@@ -9,8 +9,8 @@ import { proxyRequest, proxyRequestBare } from '../utils/RequestProxy';
 import Message from './Message';
 import Timeout = NodeJS.Timeout;
 import { LoadStatus, QueueResponse, Status } from './types';
-import { WidgetSettings } from '../../../types';
 import getWidgetSettings from '../utils/getWidgetSettings';
+import { WidgetSettings } from '../../../types';
 
 type BannerMessage = {
   messageExists: boolean;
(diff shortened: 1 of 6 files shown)


Brianzchen:
Wonder if this change should actually go ahead. An auto refresh when not in view can still be valuable so that when I have it in view again it's always up to date like a push service. 

Though given that in BB with the shitty UI I always refresh anyways cause I'm never confident with what's up to date

pastelsky:
@Brianzchen  I'd have very much liked to keep this, but for large repositories like ours and with numerous dependency checks we do start making a lot of calls to the git service and run into our API limits.

The widget already triggers an immediate refresh when it detects a change in view state as a consolation though.

Brianzchen:
Can we expose this as a flag, and we pass the flag as off? I know landkid is MOSTLY used by us but give it can be used by others we should keep that in mind.

pastelsky:
Added configurations options for tweaking refresh.

src/static/bitbucket/index.ejs
```diff
@@ -11,6 +11,9 @@
           Roboto, Helvetica, Arial, sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol";
       }
   </style>
+    <script>
+      window.__WIDGET_SETTINGS__ = <%= widgetSettings %>
```
Brianzchen, line 15: how does this work? I didn't imagine it'd be this complicated. I was imagining something like from landkid deployment for a given repo it would pass some var as a config that defaulted to some values.

This is the the way landkid handles it then it's the way it needs to be I suppose

src/static/bitbucket/components/App.tsx
```diff
@@ -71,10 +73,19 @@ const App = () => {
...
+    const isVisible =
+      !document.hidden && (widgetSettings.refreshOnlyWhenInViewport ? inViewRef.current : true);
+
+    // If only refreshing when in viewport, uses `refreshInterval`,
+    // Else uses `refreshInterval` when in viewport and twice the timeout when not in viewport
+    let refreshIntervalMs = widgetSettings.refreshOnlyWhenInViewport
+      ? widgetSettings.refreshInterval
+      : inViewRef.current
+      ? widgetSettings.refreshInterval
+      : widgetSettings.refreshInterval * 2;
```
Brianzchen, line 86: nit: just hard to read otherwise
```suggestion
    let refreshIntervalMs = widgetSettings.refreshOnlyWhenInViewport
      ? widgetSettings.refreshInterval
      : inViewRef.current
        ? widgetSettings.refreshInterval
        : widgetSettings.refreshInterval * 2;
```

src/static/bitbucket/utils/getWidgetSettings.ts
```diff
@@ -0,0 +1,10 @@
+import { WidgetSettings } from '../../../types';
+
+export default function getWidgetSettings(): WidgetSettings {
+  return {
+    refreshInterval: 30000,
```
Brianzchen, line 5: much higher than before, was 10/20 now it's 30/60 seconds

src/static/bitbucket/utils/getWidgetSettings.ts
```diff
@@ -0,0 +1,10 @@
...
+    refreshOnlyWhenInViewport: true,
```
Brianzchen, line 6: I wonder if refreshing in background as a default is more expected

src/static/bitbucket/index.ejs
pastelsky: Landkid configuration (till now) has been only used to power server-side / build time configuration though.
This is the first time we needing to pass on values from the `config.js` file onto the frontend (without going through an API call.

src/static/bitbucket/components/App.tsx
pastelsky: I tried to do the same earlier, but prettier wants to back to the version without indentation. Let me break try to break it up.

src/static/bitbucket/utils/getWidgetSettings.ts
pastelsky: Changed. FWIW "refreshing in background" only applies when the tab is in focus. We never refresh in background else.

pastelsky:
Ready to review now — CI is green.

Brianzchen (approved):
LGTM

PR #182, opened 2022-09-28
[Ready] Use an endpoint for widget settings

pastelsky:
My [previous attempt](https://github.com/atlassian/landkid/pull/181) to expose dynamic settings to the widget was un-successful. This is because `config.js` is essentially not used at build time (it is, but only for webpack dev server), and only powers the node server configuration.

Instead, I now fetch settings through a new server endpoint now. It is a bit of an overkill, but should allow us to set to tweak settings for the widget internally without needing to make and publish new landkid images in the future.

### DIFF
--- a/src/static/bitbucket/components/App.tsx
+++ b/src/static/bitbucket/components/App.tsx
@@ -9,8 +9,7 @@ import { proxyRequest, proxyRequestBare } from '../utils/RequestProxy';
 import Message from './Message';
 import Timeout = NodeJS.Timeout;
 import { LoadStatus, QueueResponse, Status } from './types';
-import getWidgetSettings from '../utils/getWidgetSettings';
-import { WidgetSettings } from '../../../types';
+import useWidgetSettings from '../utils/getWidgetSettings';
 
 type BannerMessage = {
   messageExists: boolean;
@@ -56,6 +55,7 @@ const App = () => {
   const [queue, setQueue] = useState<QueueResponse['queue'] | undefined>();
   const [_, setLoadStatus, loadStatusRef] = useState<LoadStatus>('not-loaded');
   const [state, dispatch] = useState(initialState);
+  const widgetSettings = useWidgetSettings();
 
   const { ref, inView } = useInView({
     threshold: 0,
@@ -74,7 +74,7 @@ const App = () => {
 
   // If only refreshing when in viewport, uses `refreshInterval`,
   // Else uses `refreshInterval` when in viewport and twice the timeout when not in viewport
-  const getRefreshInterval = (widgetSettings: WidgetSettings) => {
+  const getRefreshInterval = () => {
     if (widgetSettings.refreshOnlyWhenInViewport) {
       return widgetSettings.refreshInterval;
     } else {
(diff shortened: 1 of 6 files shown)


src/static/bitbucket/utils/getWidgetSettings.ts
```diff
@@ -1,10 +1,25 @@
...
+const defaultSettings = {
+  refreshInterval: 10000,
+  refreshOnlyWhenInViewport: false,
+};
+
+export default function useWidgetSettings(): WidgetSettings {
+  const [widgetSettings, setWidgetSettings] = useState<WidgetSettings>(defaultSettings);
+
+  useEffect(() => {
+    proxyRequestBare<any>('/settings', 'POST')
```
grenganathan-atlassian, line 15: any reason to not type the API response?

src/static/bitbucket/utils/getWidgetSettings.ts
```diff
@@ -1,10 +1,25 @@
...
+import { useEffect } from 'react';
 
-export default function getWidgetSettings(): WidgetSettings {
-  return {
-    refreshInterval: 10000,
-    refreshOnlyWhenInViewport: false,
-    // @ts-ignore -- this is a global variable that may be set by the server
-    ...window.__WIDGET_SETTINGS__,
-  };
+const defaultSettings = {
```
grenganathan-atlassian, line 6: The defaults could stay on the server and we can make the UI dumb by just using whatever the endpoint returns.

src/static/bitbucket/components/App.tsx
```diff
@@ -56,6 +55,12 @@ const App = () => {
   const [queue, setQueue] = useState<QueueResponse['queue'] | undefined>();
   const [_, setLoadStatus, loadStatusRef] = useState<LoadStatus>('not-loaded');
   const [state, dispatch] = useState(initialState);
+  const widgetSettings = useWidgetSettings();
+  const widgetSettingsRef = useRef(widgetSettings);
+
+  if (widgetSettings !== widgetSettingsRef.current) {
```
grenganathan-atlassian, line 61: dumb question: why is this needed?

src/static/bitbucket/utils/getWidgetSettings.ts
pastelsky: Nah, I should — copy pasta from API calls that existed before.

src/static/bitbucket/utils/getWidgetSettings.ts
pastelsky: Yeah, though there is a brief moment at the start when this setting isn't available to the client. I didn't want to block the UI on this information being available (we're fine to use defaults in the meanwhile).

src/static/bitbucket/components/App.tsx
pastelsky: This was slightly confusing to me as well (been out of touch with react for an year), but refs are needed because I'm reading these values inside a setTimeout call — which saves its own copy of variables in its closure. Using a ref (and keeping it up to date) ensures that only the latest value is used, even if we're inside a closure.

grenganathan-atlassian (approved):

PR #184, opened 2022-10-06
Clear timeouts before setting a new one

pastelsky:
Ideally given that I add a timeout to the daisy chain only after the previous one has expired, there shouldn't be timeouts running in parallel.  But I want to clear older timeouts here — just as a good practice and a remote chance that we had irregular re-renders setting up additional timeouts.


MatthewCharlton:
LGTM

MatthewCharlton (approved):

PR #187, opened 2022-10-10
Fix migration script

grenganathan-atlassian:
Faced a similar issue like - https://github.com/sequelize/sequelize/issues/2554 when deploying changes in dev.
Switched to direct query.

Tested in local and it works fine.

Will test the script again in dev and staging before deploying to prod.

### DIFF
--- a/src/db/migrations/08__mergeStrategyColumn.ts
+++ b/src/db/migrations/08__mergeStrategyColumn.ts
@@ -4,14 +4,24 @@ export default {
   up: function (query: QueryInterface) {
     return query.describeTable('LandRequest').then((table: any) => {
       if (table.mergeStrategy) return;
-      return query.sequelize.query(
-        'CREATE TYPE "enum_LandRequest_mergeStrategy" AS ENUM(\'squash\', \'merge-commit\'); ALTER TABLE "LandRequest" ADD COLUMN "mergeStrategy" "enum_LandRequest_mergeStrategy";',
-      );
+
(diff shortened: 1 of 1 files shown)


src/db/migrations/08__mergeStrategyColumn.ts
```diff
@@ -1,19 +1,27 @@
...
       if (table.mergeStrategy) return;
-      return query.addColumn('LandRequest', 'mergeStrategy', {
-        type: Sequelize.ENUM({
-          values: ['squash', 'merge-commit'],
-        }),
-        allowNull: true,
-      });
+
+      return query.sequelize
+        .query('DROP TYPE IF EXISTS "enum_LandRequest_mergeStrategy";')
```
grenganathan-atlassian, line 9: Due to a previous deployment in dev, this is needed.

pastelsky:
Ah, could we not just fix dev table instead? 
I think you should be able to execute queries on dev to undo the previous deploy.

grenganathan-atlassian:
> Ah, could we not just fix dev table instead? I think you should be able to execute queries on dev to undo the results of the previous deploy. Feels a little weird to codify this in our migration.

Tried that first and couldn't get it to work. I manually ran the 'drop type' command in dev but during deployment it created the type again but failed with 'type already exist' error without creating the column.  Some weird way in which seqelize/postgres works.

pastelsky (approved):

PR #188, opened 2022-10-12
Allow land button even when in queued state

pastelsky:
Previously, we wouldn't show a land button when the PR was in "Queued for land" state. It was expected that the Landkid poller would anyways land the PR when ready.

But it turns out that we only look at land requests in the last 7 days for polling. This means its possible for a PR to go into a "stuck" state, where it becomes "landable" after 7 days, but is still in "Queued for land".

Now ideally, after 7 days, we should mark land requests in "Queued for land" back to the default state, so this problem of stuck PRs does not arise.

But as immediate mitigation, I'm adding back the land button to the "Queued for land" in case the PR is landable.
This will also allow you to skip the 2-minute poll time if you wish and land changes immediately.

### DIFF
--- a/src/static/bitbucket/components/App.tsx
+++ b/src/static/bitbucket/components/App.tsx
@@ -227,6 +227,7 @@ const App = () => {
         queue={queue}
         status={status}
         canLandWhenAble={state.canLandWhenAble}
+        canLand={state.canLand}
         errors={state.errors}
         warnings={state.warnings}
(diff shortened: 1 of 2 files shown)


pastelsky:
The ideal thing to do here would be consider the age of the last commit in the pull request, instead of the age of the land request itself to filter through list of queued requests.

winniezheng (approved):

PR #189, opened 2022-10-12
Add merge checks error messages to reason

winniezheng:
Adding on to improving the reason messaging for failed landkid attempts. 

When we get a "Unable to merge pull request" error, there are failed merge checks fields given in the error logs. Adding this to the reason could be useful for users to know.

image

image

For long reason messages that extend the width
image

### DIFF
--- a/src/bitbucket/BitbucketAPI.ts
+++ b/src/bitbucket/BitbucketAPI.ts
@@ -9,9 +9,18 @@ import { BitbucketMerger } from './BitbucketMerger';
 
 const BBAPIBaseUrl = 'https://api.bitbucket.org/2.0/repositories';
 
+type Reason = {
+  error: {
+    fields?: {
+      merge_checks?: string[];
+    };
+    message?: string;
+  };
+};
+
 type MergePullRequestResult = {
   status: string;
-  reason?: any;
+  reason?: Reason;
 };
 
 export class BitbucketAPI {
--- a/src/bitbucket/BitbucketAPI.ts
+++ b/src/bitbucket/BitbucketAPI.ts
@@ -11,7 +11,7 @@ const BBAPIBaseUrl = 'https://api.bitbucket.org/2.0/repositories';
 
 type MergePullRequestResult = {
   status: string;
-  reason?: string;
+  reason?: any;
 };
 
 export class BitbucketAPI {
@@ -64,7 +64,7 @@ export class BitbucketAPI {
         landRequestId,
         landRequestStatus,
       });
-      return { status: BitbucketAPI.FAILED, reason: data?.error?.message || statusText };
+      return { status: BitbucketAPI.FAILED, reason: data };
     };
 
     const onMergeAbort = () => {
--- a/src/lib/Runner.ts
+++ b/src/lib/Runner.ts
@@ -270,14 +270,12 @@ export class Runner {
             targetBranch: pullRequest.targetBranch,
           });
 
-          const hasErrors = result.reason && result.reason.error;
-          const mergeCheckErrors =
-            hasErrors && result.reason.error.fields
-              ? ' ' + result.reason.error.fields.merge_checks.join(', ')
-              : '';
-          const error = hasErrors ? result.reason.error.message + '. ' + mergeCheckErrors : '';
-
-          await landRequest.setStatus('fail', 'Unable to merge pull request: ' + error);
+          await landRequest.setStatus(
+            'fail',
+            `Unable to merge pull request: ${result.reason?.error?.message}. ${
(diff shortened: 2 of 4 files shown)


src/bitbucket/BitbucketAPI.ts
```diff
@@ -11,7 +11,7 @@ const BBAPIBaseUrl = 'https://api.bitbucket.org/2.0/repositories';
 
 type MergePullRequestResult = {
   status: string;
-  reason?: string;
+  reason?: any;
```
Brianzchen, line 14: can we extend `reason` to be a union of 2 types if it's no longer just string

winniezheng: which 2 types? `reason` will be the object response returned

src/lib/Runner.ts
```diff
@@ -269,10 +269,15 @@ export class Runner {
...
-            'fail',
-            `Unable to merge pull request ${result.reason ? `: ${result.reason}` : ''}`,
-          );
+
+          const hasErrors = result.reason && result.reason.error;
+          const mergeCheckErrors =
+            hasErrors && result.reason.error.fields
+              ? ' ' + result.reason.error.fields.merge_checks.join(', ')
+              : '';
+          const error = hasErrors ? result.reason.error.message + '. ' + mergeCheckErrors : '';
```
grenganathan-atlassian, line 278: IIUC you can refactor this to something like,
```
const error = result.reason?.error?.message || result.reason?.error?.fields?.merge_checks.join(', ') || '';
```

src/bitbucket/BitbucketAPI.ts
grenganathan-atlassian: Think Brian is asking for something like 
```
reason?:  { error: { 
message?: string;
fields?: {
  merge_checks: string[]
 } 
}}
```

src/lib/Runner.ts
```diff
@@ -269,10 +269,15 @@ export class Runner {
...
           });
-          await landRequest.setStatus(
-            'fail',
-            `Unable to merge pull request ${result.reason ? `: ${result.reason}` : ''}`,
-          );
+
+          const hasErrors = result.reason && result.reason.error;
+          const mergeCheckErrors =
+            hasErrors && result.reason.error.fields
+              ? ' ' + result.reason.error.fields.merge_checks.join(', ')
```
grenganathan-atlassian, line 276: we could do `.join('\n')` so that UI does not break for lengthy strings.

winniezheng: ah that doesn't work :( I think because we're using spans... but i've made some css updates to account for lengthy strings

src/lib/Runner.ts
```diff
@@ -269,10 +269,11 @@ export class Runner {
...
           });
-          await landRequest.setStatus(
-            'fail',
-            `Unable to merge pull request ${result.reason ? `: ${result.reason}` : ''}`,
-          );
+          const error =
+            result.reason?.error?.message +
+              '. ' +
+              result.reason?.error?.fields?.merge_checks?.join(', ') ?? '';
+          await landRequest.setStatus('fail', 'Unable to merge pull request: ' + error);
```
Brianzchen, line 276: Nit: This is easier to read. String literal is much nicer to maintain cause you can read it as a real sentence instead of checking code back and forth
```suggestion
          await landRequest.setStatus(
            'fail',
            `Unable to merge pull request: ${result.reason?.error?.message}. ${result.reason?.error?.fields?.merge_checks?.join(', ') ?? ''}`,
          );
```

src/static/current-state/components/QueueItem.tsx
```diff
@@ -30,22 +30,24 @@ const queueItemStyles = css({
     textOverflow: 'ellipsis',
     whiteSpace: 'nowrap',
     overflow: 'hidden',
-    marginBottom: '3px',
+    marginBottom: '1em',
```
Brianzchen, line 33: why mix `em`?

src/lib/Runner.ts
Brianzchen: Also because it's `result.reason?.error?.message` what if anything along the optional chaining is `undefined`? We'd have a malformed error messaging

winniezheng: It would print nothing instead since we're using the nullish coalescing operator

src/static/current-state/components/QueueItem.tsx
winniezheng: oh I'm not sure, happy to just use px

Brianzchen (approved):

grenganathan-atlassian (approved):

PR #196, opened 2022-12-05
MONO-24 Updated PR priority during land-request.

GraceDmello:
**What's changed ?**
- During landing request verify if the commit build statuses for `landkid-priority` metadata.
- Increment the priority of the PR based on `landkid-priority`.

**Verfication**
Verified the changes on test repo. Need to perform e2e testing of the flow.

### DIFF
--- a/src/bitbucket/BitbucketAPI.ts
+++ b/src/bitbucket/BitbucketAPI.ts
@@ -234,7 +234,7 @@ export class BitbucketAPI {
 
     const priority =
       allBuildStatuses.find((buildStatus) => buildStatus.name === 'landkid-priority')
-        ?.description || 'HIGH';
+        ?.description || 'LOW';
 
     Logger.info('PR priority', {
       namespace: 'bitbucket:api:getPullRequestPriority',
--- a/src/bitbucket/BitbucketAPI.ts
+++ b/src/bitbucket/BitbucketAPI.ts
@@ -223,6 +223,28 @@ export class BitbucketAPI {
       }));
   };
 
+  getPullRequestPriority = async (commit: string): Promise<BB.PRPriority> => {
+    const endpoint = `${this.baseUrl}/commit/${commit}/statuses`;
+    const { data } = await axios.get<{ values: BB.BuildPriorityResponse[] }>(
+      endpoint,
+      await bitbucketAuthenticator.getAuthConfig(fromMethodAndUrl('get', endpoint)),
+    );
+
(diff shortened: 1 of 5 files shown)


src/bitbucket/BitbucketAPI.ts
```diff
@@ -223,6 +223,28 @@ export class BitbucketAPI {
...
+    const { data } = await axios.get<{ values: BB.BuildPriorityResponse[] }>(
+      endpoint,
+      await bitbucketAuthenticator.getAuthConfig(fromMethodAndUrl('get', endpoint)),
+    );
+
+    const allBuildStatuses = data.values;
+
+    const priority =
+      allBuildStatuses.find((buildStatus) => buildStatus.name === 'landkid-priority')
+        ?.description || 'HIGH';
```
Blasz, line 237: Should the default value here be `LOW` instead of `HIGH`?

We could also validate the `description` here as well if we field as well to ensure its one of `LOW` or `HIGH` but that may be overkill.

src/routes/bitbucket/proxy/index.ts
```diff
@@ -180,6 +181,17 @@ export function proxyRoutes(runner: Runner, client: BitbucketClient) {
...
+      const priority = await client.bitbucket.getPullRequestPriority(commit);
+
+      if (priority === 'HIGH' && request) {
+        await request.incrementPriority();
+        Logger.info('Pull request priority increased', {
+          namespace: 'routes:bitbucket:proxy:land',
+          pullRequestId: prId,
+          landRequest,
+        });
+      }
```
Blasz, line 193: An alternative to incrementing priority after queueing would be to set the priority on land request creation. This works though.

Blasz (changes requested):
Looks good, only blocker is the default value. Looks like builds weren't triggered on the PR because its from a fork. I've updated the project settings to run builds on forks so it should run on your next push.

src/bitbucket/BitbucketAPI.ts
GraceDmello: Yes, the default value should be `LOW`. Missed to revert the changes after testing.

GraceDmello: It feels an overkill at this stage. Can be added if we plan to send more meta-data.

Blasz (approved):

Blasz (approved):

PR #197, opened 2022-12-05
Add duration fields to LAND_REQUEST.STATUS.CHANGED event

Blasz:
These can be used to track how long each status transition takes as well as how long it has been since the request was originally queued.

### DIFF
--- a/src/db/index.ts
+++ b/src/db/index.ts
@@ -148,6 +148,7 @@ export class LandRequest extends Model<LandRequest> implements ILandRequest {
       prevState: prevStatus?.state,
       stateDuration: prevStatus ? Date.now() - prevStatus.date.getTime() : null,
       durationSinceQueued: queuedDate ? Date.now() - queuedDate.getTime() : null,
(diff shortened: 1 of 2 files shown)


src/db/index.ts
```diff
@@ -148,6 +146,8 @@ export class LandRequest extends Model<LandRequest> implements ILandRequest {
       state,
       reason,
       prevState: prevStatus?.state,
+      stateDuration: prevStatus ? Date.now() - prevStatus.date.getTime() : null,
+      durationSinceQueued: queuedDate ? Date.now() - queuedDate.getTime() : null,
```
GraceDmello, line 150: durationSinceQueued should give cumulative queue time when LandRequestStatus `state` is `merging`?

GraceDmello (approved):

src/db/index.ts
Blasz: Yep, although `success` is slightly more accurate

GraceDmello (approved):

PR #198, opened 2023-01-04
MONO-110 Abort 'land when able' requests if PR commit or target branch has changed

Blasz:
This solves the problem described in MONO-110 in a slightly different way. Instead of updating landkid to use the latest PR commit when transitioning from queue -> running, we abort the 'land when able' request as soon as a new commit has been pushed.

This is beneficial for a few reasons:

1. Provides more control to the PR author over how they want to land their changes by giving them a way to cancel a 'land when able' request. They may not want to land when able after pushing new commits.
2. Solves the problem of new commits being pushed after 'land when able' (but before queued) causing unavoidable land failures
3. Keeps the system robust by enforcing that the PR commit at the time of queueing is the same at land time

This approach does not support the case whereby an author queues a PR, pushes a new commit that then passes all land checks in time before reaching the front of the queue. However, I don't think we need or should support that use case. It's an edge case and I don't think it's unreasonable for a developer to re-queue their PR after pushing new changes.

Ideally we'd be able to de-queue PRs already in the queue after pushing new commits, which is something we might explore later as part of the conflict and queue health analyser in M5.

I've added tests for some of the land request transition behaviour (there were none previously). It's still not in an ideal state but it's a start. Ultimately I think we'd want to mock out the database with an in-memory equivalent so we can easily unit test transitions without manually mocking everything.


GraceDmello:
Based on GitHub auto-merge feature, more commits can be pushed. Land when able seems to be similar feature and should support adding more commits.

The test coverage for the runner is very limited and it's very risky to make changes without them. This is a good start for that.

Blasz:
As discussed offline, we'll keep the more restrictive approach for now where we cancel auto-merging when new commits are pushed since we don't have a way to toggle off auto-merging. We'll raise another task to toggle this behaviour and can revisit supporting auto merging subsequent commits/pushes then.

GraceDmello (approved):

PR #202, opened 2023-01-10
Added timeout for long running landkid builds.

GraceDmello:
Landkid is been hung multiple times now, where in the status of the running land request is not updated. This causes the running land request to be not updated. 

One possible root cause of the issue is  the ‘build finished’ webhook is not processed successfully.

The original ticket mentions auto landing the request. However, failing the hung request which breaches the timeout period should also be sufficient to break the loop.

image


src/lib/Runner.ts
```diff
@@ -46,6 +48,11 @@ export class Runner {
     setInterval(() => {
       this.next();
     }, 15 * 1000); // 15s
+
+    // call checkRunningLandRequests() function on a interval of 10 min to verify LAND_BUILD_TIMEOUT_TIME
+    setInterval(() => {
+      this.checkRunningLandRequests();
+    }, 10 * 60 * 1000);
```
GraceDmello, line 55: Should the interval of 10 min sufficient ? The scenario of hung system is quite rare. So, don't want to overload the system with too many checks.

src/lib/Runner.ts
```diff
@@ -25,6 +25,8 @@ import { BitbucketAPI } from '../bitbucket/BitbucketAPI';
 // 3 hours should be long enough to complete it now that we only fetch requests within 7 days (about 20 waiting requests)
 const MAX_CHECK_WAITING_REQUESTS_TIME = 1000 * 60 * 60 * 3; // 3 hours
 
+const LAND_BUILD_TIMEOUT_TIME = 1000 * 60 * 60 * 2; // 2 hours
```
GraceDmello, line 28: Should this be configurable or static value would do as well ?

src/bitbucket/__tests__/BitbucketPipelinesAPI.test.ts
```diff
@@ -0,0 +1,37 @@
...
+    getAuthConfig: () => Promise.resolve({}),
+  },
+}));
+
+const bitbucketPipelineAPI = new BitbucketPipelinesAPI({
+  repoName: 'repo',
+  repoOwner: 'owner',
+});
+
+describe('BitbucketPipelinesAPI', () => {
```
Blasz, line 18: We should add a `beforeEach` that clears/resets mocks to ensure they don't leak between tests. Saves having to refactor it when we add a test for next time.

You might have to re-apply the `BitbucketAuthenticator` mock in the beforeEach if you reset the mocks. Otherwise you'll need to update ` mockedAxios.get.mockResolvedValue(response)` to only mock it once or manually reset that mock so it doesn't persist across tests.

src/bitbucket/__tests__/BitbucketPipelinesAPI.test.ts
```diff
@@ -0,0 +1,37 @@
+import axios from 'axios';
+import { BitbucketPipelinesAPI } from '../BitbucketPipelinesAPI';
+
+jest.mock('axios');
+const mockedAxios = axios as unknown as jest.Mocked<typeof axios>;
+
+jest.mock('../BitbucketAuthenticator', () => ({
+  bitbucketAuthenticator: {
+    getAuthConfig: () => Promise.resolve({}),
```
Blasz, line 9: generally a good practice to use `jest.fn()` so the mock implementation can be changed between tests.

e.g.

```
getAuthConfig: jest.fn(() => Promise.resolve({}))
```

Can also use make the fn `async` to get rid of the `Promise.resolve`

src/lib/Runner.ts
Blasz: I think static is fine

src/lib/Runner.ts
Blasz: Yep I think 10 minutes is good

src/lib/Runner.ts
```diff
@@ -630,6 +636,57 @@ export class Runner {
...
+                await landRequest.setStatus('fail', 'Build timeout period breached');
+              } else {
+                const { buildId } = landRequestStatus.request;
+                const { state } = await this.client.getLandBuild(buildId);
+
+                // buildStatus can be SUCCESSFUL, FAILED or STOPPED
+                this.onStatusUpdate({
+                  buildId,
+                  buildStatus: state.result?.name,
+                });
```
Blasz, line 677: I think we should be `await`ing the `onStatusUpdate` call since its async

Blasz (commented):
Nice one. Just a few comments.

src/bitbucket/__tests__/BitbucketPipelinesAPI.test.ts
GraceDmello: Yup, I have rested the mocks in `beforeEach` and moved the mocks for `BitbucketAuthenticator ` within `beforeEach`

src/bitbucket/__tests__/BitbucketPipelinesAPI.test.ts
GraceDmello: Changed the mocking to use `mockResolvedValue`.

src/lib/Runner.ts
GraceDmello: Yeah, I realised this from the logging as well while testing. Added the change.

Blasz (approved):

PR #207, opened 2023-02-01
MONO-301 Land priority branch UI implementation

winniezheng:
Added priority branch UI to admin controls
- Priority branches can be added, edited, removed
image

- Duplication check on edit/add
image
image

Example use where `test-priority/*` has been added to the priority branch list (priority has been bumped for priority branch)
image

**Test plan**
- added unit tests
- integration tests to be added in the `atlassian-frontend-landkid-deployment` repo 
- tested against scenarios where multiple priority branches are in the queue - the priority will be given to the priority branch queued first (I think this is fine but we could possibly look at "ordering" the priority in the future)

### DIFF
--- a/src/routes/bitbucket/proxy/index.ts
+++ b/src/routes/bitbucket/proxy/index.ts
@@ -8,6 +8,7 @@ import { Runner } from '../../../lib/Runner';
 import { Logger } from '../../../lib/Logger';
 import { eventEmitter } from '../../../lib/Events';
 import { StateService } from '../../../lib/StateService';
+import { validatePriorityBranch } from '../../../lib/utils/helper-functions';
 
 interface LandBody {
   mergeStrategy?: IMergeStrategy;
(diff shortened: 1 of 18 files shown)


vctr-dev (dismissed):

src/routes/bitbucket/proxy/index.ts
```diff
@@ -182,6 +183,17 @@ export function proxyRoutes(runner: Runner, client: BitbucketClient) {
         targetBranch: prInfo.targetBranch,
       });
 
+      const priorityBranches = await StateService.getPriorityBranches();
+      const isPriorityBranch = validatePriorityBranch(priorityBranches, landRequest.prSourceBranch);
+      if (isPriorityBranch && request) {
+        await request.incrementPriority();
```
cheesehary, line 189: This increments the priority of the request by 1 when landing. I guess it works assuming no other land requests in the queue are prioritised manually. What if there already are requests with priority modified in the queue? Or we are ok to give priority to those requests

cheesehary (commented):
Nice one

src/routes/bitbucket/proxy/index.ts
winniezheng: That's a good point - if land requests in the queue are manually prioritised first, then any other requests prioritised after will be behind them, but still prioritised before other requests. I feel like it's a bit of an edge case though, it's not often branches are manually prioritised and if they are they are, they're probably more urgent and one-offs?

cheesehary: yeah fair enough

cheesehary (dismissed):

cheesehary (approved):

PR #210, opened 2023-02-10
Retry failed PR merges

Blasz:
Will retry failed merges twice in 15s intervals.

PR #211, opened 2023-02-13
New API route updates to assist with e2e tests

Blasz:
I've been running into flakey cypress tests when trying to integrate the changes from https://github.com/atlassian/landkid/pull/210 into atlassian-frontend.

Some of the flakiness relates to server state not being reset in between tests/test suites/test executions - especially if there is a failure.

I've made a couple of changes to help with this:

* New route to delete all priority branches
* Update to the recently added routes to return error status codes on failure. I noticed these were not awaiting the DB calls and always returning success.

I plan to update the cypress test code to these routes to reset server state where required.

### DIFF
--- a/src/routes/api/index.ts
+++ b/src/routes/api/index.ts
@@ -149,8 +149,12 @@ export function apiRoutes(runner: Runner, client: BitbucketClient, config: Confi
       });
 
       if (typeof maxConcurrentBuilds == 'number' && maxConcurrentBuilds > 0) {
-        StateService.updateMaxConcurrentBuild(maxConcurrentBuilds, req.user!);
-        return res.status(200).json({ success: true });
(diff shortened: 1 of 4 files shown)


winniezheng (approved):

src/routes/api/index.ts
```diff
@@ -208,8 +220,28 @@ export function apiRoutes(runner: Runner, client: BitbucketClient, config: Confi
...
+      }
+    }),
+  );
+
+  router.post(
+    '/remove-all-priority-branches',
+    requireAuth('admin'),
+    wrap(async (req, res) => {
+      Logger.verbose(`Removing all priority branches`, {
+        namespace: 'routes:api:remove-priority-branch',
```
GraceDmello, line 237: nit: Should the namespace be `routes:api:remove-all-priority-branches`

GraceDmello (approved):

PR #212, opened 2023-02-15
Resolve unexpected land request state changes

GraceDmello:
**Problem**
There are few instances of unexpected state transition. The newly introduced `checkRunningLandRequests` triggers `onStatusUpdate` operation. `onStatusUpdate` is also triggered via webhook when land build is completed. The race condition between this two operation causes this unexpected transition.

 
**Solution**

Updated `onStatusUpdate` operation to consider land request status which are in running state. Previously, `awaiting-merge` and `merging` statuses were also considered.

image

### DIFF
--- a/src/lib/Runner.ts
+++ b/src/lib/Runner.ts
@@ -360,52 +360,46 @@ export class Runner {
 
   // onStatusUpdate only updates status' for landrequests, never moves a state and always exits early
   onStatusUpdate = async (statusEvent: BB.BuildStatusEvent) => {
-    const running = await this.getRunning();
-    if (!running.length) {
-      Logger.info('No builds running, status event is irrelevant', {
+    const requests = await this.getRunning();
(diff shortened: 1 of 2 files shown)


src/lib/Runner.ts
```diff
@@ -360,52 +360,46 @@ export class Runner {
...
+        logMessage('Moving landRequest to failed state');
+        await landRequest.setStatus('fail', 'Landkid build failed');
+        return this.next();
+      case 'STOPPED':
+        logMessage('Moving landRequest to aborted state');
+        await landRequest.setStatus('aborted', 'Landkid pipelines build was stopped');
+        return this.next();
+      default:
+        logMessage('Dont know what to do with build status, ignoring');
+        break;
```
Blasz, line 402: Much cleaner 😄

Blasz (approved):

PR #219, opened 2023-03-05
Initial implementation of speculation engine

GraceDmello:
Added speculation engine to re-order PR based on impact. It is target repos's responsibility to provide impact of each PR using BBP build statuses. The engine attempts to re-order a land-request if

1. speculationEngineEnabled feature is enabled
2. available free running slots are at least 2
3. Position of current requests in the queue compared to number of slots.
            for example if 2 slots are available consider re-ordering 1st request from the queue
4. Size of queue is more than 1
 5. Impact of the current request compared to the next in queue

The feature can be live toggled using below admin controls.
image

### DIFF
--- a/src/lib/Runner.ts
+++ b/src/lib/Runner.ts
@@ -12,6 +12,7 @@ import { eventEmitter } from './Events';
 import { BitbucketAPI } from '../bitbucket/BitbucketAPI';
 import { StateService } from './StateService';
 import { validatePriorityBranch } from './utils/helper-functions';
+import { SepculationEngine } from './SpeculationEngine';
 
 // const MAX_WAITING_TIME_FOR_PR_MS = 2 * 24 * 60 * 60 * 1000; // 2 days - max time build can "land-when able"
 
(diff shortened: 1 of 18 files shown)


src/lib/Runner.ts
```diff
@@ -70,6 +71,12 @@ export class Runner {
     return running.length >= maxConcurrentBuilds;
   };
 
+  getAvailableSlots = async (processingLandRequestStatuses: LandRequestStatus[]) => {
```
GraceDmello, line 74: remove this

src/lib/SpeculationEngine.ts
```diff
@@ -0,0 +1,95 @@
+import { LandRequest, LandRequestStatus } from '../db';
+import { Logger } from './Logger';
+import { StateService } from './StateService';
+
+export class SepculationEngine {
```
winniezheng, line 5: Nit: typo here "SepculationEngine" (and all the imported references)

src/lib/SpeculationEngine.ts
```diff
@@ -0,0 +1,95 @@
...
+      });
+
+    logMessage('Speculation engine details', {
+      availableSlots,
+      position,
+      queued: queued.map(({ request }) => request.pullRequestId),
+    });
+
+    if (availableSlots < 2 || queued.length < 2 || position >= availableSlots - 1) {
+      logMessage('Not attempting to re-order the PR');
```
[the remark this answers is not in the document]
GraceDmello: will update this on the integration PR.

PR #221, opened 2023-03-08
Integrate speculation engine with impact

GraceDmello:
Added speculation engine to re-order PR based on impact. It is target repos's responsibility to provide impact of each PR using BBP build statuses. The engine attempts to re-order a land-request if

1. speculationEngineEnabled feature is enabled
2. available free running slots are at least 2
3. Position of current requests in the queue compared to number of slots.
            for example if 2 slots are available consider re-ordering 1st request from the queue
4. Size of queue is more than 1
 5. Impact of the current request compared to the next (available slots - 1) number of requests in queue

The feature can be live toggled using below admin controls.
image

Impact retrieved and integrated for re-ordering.

image

image

Impact displayed on the UI
image

### DIFF
--- a/src/lib/SpeculationEngine.ts
+++ b/src/lib/SpeculationEngine.ts
@@ -2,7 +2,7 @@ import { LandRequest, LandRequestStatus } from '../db';
 import { Logger } from './Logger';
 import { StateService } from './StateService';
 
-export class SepculationEngine {
+export class SpeculationEngine {
   constructor() {}
 
   static async getAvailableSlots(running: LandRequestStatus[]): Promise<number> {
--- a/src/lib/SpeculationEngine.ts
+++ b/src/lib/SpeculationEngine.ts
@@ -17,9 +17,9 @@ export class SepculationEngine {
   static async getImpact(queued: LandRequestStatus[], position: number) {
     const currentRequestStatus = queued[position];
     const nextRequestStatus = queued[position + 1];
-    // TODO: Replace with impact from land request.
-    const currentImpact = Math.round(Math.random() * 100);
-    const nextImpact = Math.round(Math.random() * 100);
+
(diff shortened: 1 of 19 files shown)


src/lib/SpeculationEngine.ts
```diff
@@ -14,22 +14,21 @@ export class SpeculationEngine {
...
+  static async getImpact(queued: LandRequestStatus[], position: number, availableSlots: number) {
+    // Compare current request with availableSlots - 1 number of next queue items
+    const queuedRequestsLength = position + availableSlots - 1;
+    const nextQueuedRequestStatus = queued.slice(position, queuedRequestsLength + 1);
 
-    const currentImpact = currentRequestStatus.request.impact;
-    const nextImpact = nextRequestStatus.request.impact;
+    const queuedRequestImpacts = nextQueuedRequestStatus.map((requestStatus) => {
+      return requestStatus.request.impact;
+    });
```
GraceDmello, line 24: nit:
```
const queuedRequestImpacts = nextQueuedRequestStatus.map(({request}) => request.impact);
```

src/lib/SpeculationEngine.ts
```diff
@@ -14,22 +14,21 @@ export class SpeculationEngine {
...
 
-    Logger.info('Impact retrieved:', {
+    Logger.info('Impact retrieved for next queued requests:', {
       namespace: 'lib:speculationEngine:getImpact',
-      pullRequestId: currentRequestStatus.request.pullRequestId,
-      nextPullRequestId: nextRequestStatus.request.pullRequestId,
-      currentImpact,
-      nextImpact,
+      impact: queuedRequestImpacts,
```
GraceDmello, line 28: can we also add `pullRequestId` to the logging for the current PR. It will be useful for debugging via Splunk.

winniezheng (approved):

PR #223, opened 2023-03-09
Added speculation engine 

GraceDmello:
Added speculation engine to re-order PR based on impact. It is the target repos' responsibility to provide the impact of each PR using BBP build statuses. The engine attempts to re-order a land-request if

- `speculationEngineEnabled` feature is enabled
- Available free running slots are at least 2
- Position of current requests in the queue compared to number of slots. For example:
if 2 slots are available consider re-ordering 1st request from the queue;
- Size of queue is more than 1
- Impact of the current request compared to the next (available slots - 1) number of requests in queue

Used BBC APIs to create a build status

image

Add impact field to LandRequest

image

Impact displayed on the UI

image

The feature can be live toggled using below admin controls

image


src/lib/__tests__/SepculationEngine.test.ts
```diff
@@ -0,0 +1,254 @@
...
+      const reOrder = await SpeculationEngine.reOrderRequest(
+        mockRunning,
+        mockQueued,
+        mockQueued[0],
+      );
+
+      expect(SpeculationEngine.getImpact).not.toHaveBeenCalled();
+      expect(reOrder).toBe(false);
+    });
+    describe('when feature is turned on', () => {
```
Blasz, line 131: I think it'd be better to use these `describe` blocks to distinguish between successful and unsuccessful reorder cases. It reduces cognitive overhead and means we can more easily identify missing tests such as there not being a test for a successful re-order with 3 free slots.

The feature being turned off would just be a 'unsuccessful reorder case', it doesn't warrant being its own describe imo.

src/lib/__tests__/SepculationEngine.test.ts
```diff
@@ -0,0 +1,254 @@
+import { LandRequest, LandRequestStatus, PullRequest } from '../../db';
```
Blasz, line 1: The filename has a typo

Blasz (changes requested):
This looks super cool, great work folks 👏.

Most of the comments are nits but there are a couple that I think should definitely be addressed. Biggest one is the inconsistency between whether higher or lower impact scores are given priority.

config.example.js
```diff
@@ -76,6 +76,11 @@ module.exports = {
       ],
     },
   },
+  /**
+   * Speculation engine re-orders top n PRs where n is the available free slots based on the impact of the PRs. The lower impact PRs are given preference.
+   * Impact meta data is processed and send to landkid by the consuming repo using build statuses.
+   */
+  speculationEngineEnabled: false,
```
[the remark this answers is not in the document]
GraceDmello: will nest `speculationEngineEnabled` within new `queueSettings` field.

src/db/migrations/09__impactColumn.ts
```diff
@@ -0,0 +1,18 @@
...
+        allowNull: true,
+        defaultValue: 0,
+      });
+    });
+  },
+  // VIOLATES FOREIGN KEY CONSTRAINT
+  down: function () {
+    throw new Error('NO DROP FUNCTION FOR THIS MIGRATION');
+  },
+};
```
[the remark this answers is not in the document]
winniezheng: I believe it's needed. I ran into this error on creating a new column to an existing table: `error: [lib:utils:locker:withLock] Error failed while in lock for "status-transition" {"errString":"SequelizeDatabaseError: SQLITE_ERROR: no such column: request.test","errStack":"SequelizeDatabaseError: SQLITE_ERROR: no such column: request.test\n  `

src/lib/SpeculationEngine.ts
```diff
@@ -0,0 +1,105 @@
...
+   *  3. Size of the queue is greater than 1
+   *  4. Impact of the current request is not the lowest impact compared with the requests behind it in the queue
+   *
+   * @param running
+   * @param queued
+   * @param currentLandRequestStatus
+   * @returns true if requests needs to be reordered.
+   *
+   */
+  static async reOrderRequest(
```
GraceDmello, line 61: nit: reorder is one word so I don't think O should be capitalised here

src/lib/SpeculationEngine.ts
```diff
@@ -14,54 +14,63 @@ export class SpeculationEngine {
...
   }
 
-  static async getImpact(queued: LandRequestStatus[], position: number, availableSlots: number) {
-    // Compare current request with availableSlots - 1 number of next queue items
-    const queuedRequestsLength = position + availableSlots - 1;
-    const nextQueuedRequestStatus = queued.slice(position, queuedRequestsLength + 1);
+  // get the request with the lowest impact by comparing the current request with the next requests behind in the queue
+  static getLowestImpact(
+    queued: LandRequestStatus[],
+    position: number,
```
GraceDmello, line 20: `position` is no longer needed now.

src/lib/SpeculationEngine.ts
```diff
@@ -14,54 +14,63 @@ export class SpeculationEngine {
...
+    availableSlots: number,
+  ): LandRequestStatus {
+    // the number of queued requests we want to compare the impact with should equal to the number of available slots
+    const nextQueuedRequestStatus = queued.slice(0, availableSlots);
 
-    const queuedRequestImpacts = nextQueuedRequestStatus.map(
-      (requestStatus) => requestStatus.request.impact,
+    // sort the impact of next queued requests so that we can compare the lowest request's impact with the current request's impact
+    const queuedRequestImpacts = nextQueuedRequestStatus.sort(
```
GraceDmello, line 27: the var name can be `sortedRequestStatuses`

src/lib/SpeculationEngine.ts
```diff
@@ -14,54 +14,63 @@ export class SpeculationEngine {
...
+      (reqStatusA, reqStatusB) => reqStatusA.request.impact - reqStatusB.request.impact,
     );
 
     Logger.info('Impact retrieved for next queued requests:', {
       namespace: 'lib:speculationEngine:getImpact',
       pullRequestId: queued[position].request.pullRequestId,
       impact: queuedRequestImpacts,
```
GraceDmello, line 34: `sortedRequestStatuses: queuedRequestImpacts.map(({ request }) => request.pullRequestId + ' ' + request.impact)`

src/lib/SpeculationEngine.ts
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    logMessage('Attempting to re-order PR based on impact');
+    // getImpactQueuedRequests will return an array of impact values from the current request and the next number (available slots - 1) of requests in the queue
+    const getImpactQueuedRequests = await this.getImpact(queued, position, availableSlots);
+    const getImpactQueuedRequestsSorted = [...getImpactQueuedRequests].sort();
+
+    //if our current request has a lower impact than the next queued requests, then we should reorder this request
+    if (getImpactQueuedRequests[position] !== getImpactQueuedRequestsSorted[position]) {
+      // re-order as current impact is greater than next impact
```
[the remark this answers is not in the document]
GraceDmello: The lower impact PRs should be ordered first. Comments and logic is updated to address it.

src/lib/Runner.ts
```diff
@@ -128,6 +129,16 @@ export class Runner {
       return landRequest.setStatus('fail', 'Unable to land due to failed land checks');
     }
 
+    if (
+      await SpeculationEngine.reorderRequest(
+        runningTargetingSameBranch,
+        await this.getQueue(['queued']),
```
Blasz, line 135: Just realised something. Should we also be filtering the queue to the same target branch?

src/lib/SpeculationEngine.ts
```diff
@@ -0,0 +1,117 @@
+import { LandRequest, LandRequestStatus } from '../db';
+import { Logger } from './Logger';
+import { StateService } from './StateService';
+
+/**
+ *  SpeculationEngine uses impact of a PR to reorder it while placing the PRs on the running slots.
+ *  The lower impact PRs are given higher precedence.
+ *  Impact meta data is processed and send to landkid by the consuming repo using build statuses.
+ * */
```
Blasz, line 9: Nit: Might be worth explaining why this is beneficial. I.e. to reduce the likelihood of a landkid failures due to dependencies

src/lib/SpeculationEngine.ts
```diff
@@ -0,0 +1,117 @@
...
+
+    logMessage('Attempting to reorder PR based on impact');
+    const lowestImpact = this.getLowestImpactedRequestStatus(
+      queue,
+      positionInQueue,
+      availableSlots,
+    );
+
+    // if the curent request has the lowest impact, we do not need to reorder this request.
+    if (queue[positionInQueue].request.id === lowestImpact.request.id) {
```
Blasz, line 110: nit: `queue[positionInQueue].request` is the same as `landRequest` right? I think referencing the latter is clearer.

src/lib/__tests__/SpeculationEngine.test.ts
```diff
@@ -0,0 +1,284 @@
...
+      );
+
+      expect(SpeculationEngine.getLowestImpactedRequestStatus).toHaveBeenCalledWith(
+        mockQueued.slice(0, 2),
+        0,
+        2,
+      );
+      expect(reorder).toBe(true);
+    });
+  });
```
Blasz, line 149: Do we want to add a successful case for 3 slots as well?

Blasz (commented):
Looks good. Just the one comment on queue filtering.

src/lib/Runner.ts
GraceDmello: good catch. Yes, we should be filtering the queue to the same target branch.

src/lib/__tests__/SpeculationEngine.test.ts
GraceDmello: Yeah, can add another test case.

src/lib/Runner.ts
```diff
@@ -761,7 +774,13 @@ export class Runner {
     });
     return landRequestStatus;
   };
-
+  // filter queue by target branch - used by speculation engine to reorder PRs by target branch
+  filterQueue = async (landRequest: LandRequest) => {
```
GraceDmello, line 778: nit: can be renamed to `filterQueueByTargetBranch` and the input of the function can be `targetBranch`

Blasz (approved):

PR #228, opened 2023-03-29
Fix issues when there are too many PR dependencies

Blasz:
PRs that transition to running with more than 7 dependencies (including both running and awaiting-merge) will cause a `SequelizeDatabaseError: value too long for type character varying(255)` sequelize error.
This is caused by  dependsOn column of a landRequest table exceeding the max string length of 255 characters. The dependsOn column contains each dependency landRequest ID (36 characters) separated by commas (1 column). So only `Math.floor(255/36) = 6` dependencies can fit in the column. This means 2 running requests + 4 PRs awaiting merge.

When the sequelize error occurs, neither the dependsOn or buildId fields are attached to the land request after its status is set to running. This then results in the request eventually timing out after 2 hours due to our landkid build timeout handling that we added recently.

I've added the following to rectify this issue:

* Update the max length of the dependsOn column from 255 chars to 1000 chars
* Include a hard limit of dependencies of 25 so that we do not exceed the 1000 char limit
* Handle transitions to running more robustly by catching these types of errors and setting the status to failed to remove them from the queue immediately.
* Add an additional check for buildId in our timeout check that will fail the request if no buildId is found

### DIFF
--- a/src/db/migrations/09__updateDependsOnColumn.ts
+++ b/src/db/migrations/09__updateDependsOnColumn.ts
@@ -1,0 +1,16 @@
+import { QueryInterface, DataTypes } from 'sequelize';
+
+export default {
+  up: async function (query: QueryInterface, Sequelize: DataTypes) {
+    return query.changeColumn('LandRequest', 'dependsOn', {
+      type: Sequelize.STRING({ length: 1000 }), // set the maximum length to 1000
+      allowNull: true,
+    });
+  },
+  down: async function (query: QueryInterface, Sequelize: DataTypes) {
+    return query.changeColumn('LandRequest', 'dependsOn', {
+      type: Sequelize.STRING(), // set the maximum length back to default 255
+      allowNull: true,
+    });
+  },
+};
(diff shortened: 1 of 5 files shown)


src/db/migrations/09__updateDependsOnColumn.ts
```diff
@@ -0,0 +1,16 @@
...
+    return query.changeColumn('LandRequest', 'dependsOn', {
+      type: Sequelize.STRING({ length: 1000 }), // set the maximum length to 1000
+      allowNull: true,
+    });
+  },
+  down: async function (query: QueryInterface, Sequelize: DataTypes) {
+    return query.changeColumn('LandRequest', 'dependsOn', {
+      type: Sequelize.STRING(), // set the maximum length back to default 255
+      allowNull: true,
+    });
```
GraceDmello, line 14: Why do we need to revert back?

GraceDmello (approved):

src/db/migrations/09__updateDependsOnColumn.ts
Blasz: We shouldn't, it's just in case we need to rollback for whatever reason.

src/lib/Runner.ts
```diff
@@ -65,6 +69,18 @@ export class Runner {
   };
 
   areMaxConcurrentBuildsRunning = async (processingLandRequestStatuses: LandRequestStatus[]) => {
+    if (processingLandRequestStatuses.length > MAX_LAND_DEPENDENCIES) {
+      // Set a hard limit on the total number of running + awaiting-merge + merge
+      Logger.warn('Hit max number of landkid dependencies', {
```
winniezheng, line 74: Does this mean that the current request will just remain in the queue until the next "dependency slot" is free?

PR #230, opened 2023-04-12
MONO-521 Added landBuildTimeoutTime to instance configuration

GraceDmello:
Added landBuildTimeoutTime to system configuration. So, each instance can update the timeout based on their CI requirements. It defaults to 120 mins if not configured.

### DIFF
--- a/src/bitbucket/BitbucketAPI.ts
+++ b/src/bitbucket/BitbucketAPI.ts
@@ -42,7 +42,7 @@ export class BitbucketAPI {
     options: MergeOptions = {},
   ): Promise<MergePullRequestResult> => {
     const {
-      id: landRequestId,
+      requestId: landRequestId,
       request: {
         pullRequestId,
         pullRequest: { targetBranch },
--- a/src/lib/__tests__/Runner.test.ts
+++ b/src/lib/__tests__/Runner.test.ts
@@ -77,6 +77,7 @@ describe('Runner', () => {
   };
   let mockPullRequest: BB.PullRequest;
   let mockClient: BitbucketClient;
+  let mockConfig: Config;
 
   beforeEach(() => {
     mockQueue = {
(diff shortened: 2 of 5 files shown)


src/bitbucket/BitbucketAPI.ts
```diff
@@ -42,7 +42,7 @@ export class BitbucketAPI {
     options: MergeOptions = {},
   ): Promise<MergePullRequestResult> => {
     const {
-      id: landRequestId,
+      requestId: landRequestId,
```
GraceDmello, line 45: This caused issues while debugging as land request status id was printed instead of land request id.

src/lib/__tests__/Runner.test.ts
```diff
@@ -439,6 +443,23 @@ describe('Runner', () => {
       );
     });
 
+    test('should fail land request if timeout period configured via config is breached', async () => {
+      mockConfig.prSettings.landBuildTimeoutTime = 1000 * 60 * 60;
+      //running state is beyond the timeout period of 2 hours
```
cheesehary, line 448: nit: I think it's 1 hour in this test case

src/lib/__tests__/Runner.test.ts
```diff
@@ -439,6 +443,23 @@ describe('Runner', () => {
...
+      const mockLandRequestStatus = getLandRequestStatus(new Date('2022-12-13T03:42:48.071Z'));
```
cheesehary, line 449: I'm curious why this specific date?

cheesehary (approved):

src/lib/__tests__/Runner.test.ts
GraceDmello: The code looks for `current date - this date` to calculated the time elapsed. We haven't mocked the current date. So, any timestamp in past should be sufficient to test the scenario

PR #235, opened 2023-06-15
Fix/dependencies bug

winniezheng:
This PR fixes the "??" bug which was introduced after the new merge queue was implemented. When the dependent builds are placed in the merging queue, the dependent builds are no longer in the queue. The logic to find the dependent PRs cannot find these request ids, so it defaults to "??"

Changes in this PR:
- Added dependsOnPrIds column to the LandRequest table 
- Stores prIds as dependencies in `moveFromQueueToRunning`  
- Renders dependsOnPrIds in `QueueItem` component

Tested in staging ✅ 
image

### DIFF
--- a/src/db/migrations/11__dependsOnPrIds.ts
+++ b/src/db/migrations/11__dependsOnPrIds.ts
@@ -2,13 +2,13 @@ import { QueryInterface, DataTypes } from 'sequelize';
 
 export default {
   up: async function (query: QueryInterface, Sequelize: DataTypes) {
-    return query.changeColumn('LandRequest', 'dependsOnPrIds', {
+    return query.addColumn('LandRequest', 'dependsOnPrIds', {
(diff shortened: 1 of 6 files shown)


src/db/migrations/11__dependsOnPrIds.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export default {
+  up: async function (query: QueryInterface, Sequelize: DataTypes) {
+    return query.addColumn('LandRequest', 'dependsOnPrIds', {
+      type: Sequelize.STRING({ length: 1000 }), // set the maximum length to 1000
+      allowNull: true,
+    });
+  },
+  down: async function (query: QueryInterface, Sequelize: DataTypes) {
+    return query.addColumn('LandRequest', 'dependsOnPrIds', {
```
GraceDmello, line 11: Do we need migration for down operation ?

GraceDmello (approved):

PR #236, opened 2023-06-19
[dev] Fail all dependents in one go.

GraceDmello:
Updated the BBC webhook callback utility to fail all dependent requests simultaneously. This would create more scenarios where multiple running slots are freed up, eventually causing the speculation engine to kick in for re-ordering pull requests based on impact.

### DIFF
--- a/src/lib/Runner.ts
+++ b/src/lib/Runner.ts
@@ -99,8 +99,6 @@ export class Runner {
         namespace: 'lib:runner:moveFromQueueToRunning',
         landRequestId: landRequest.id,
         pullRequestId: landRequest.pullRequestId,
-        // landRequestStatus,
-        // runningTargetingSameBranch,
         lockId,
       });
       return false;
--- a/src/lib/Runner.ts
+++ b/src/lib/Runner.ts
@@ -99,8 +99,8 @@ export class Runner {
         namespace: 'lib:runner:moveFromQueueToRunning',
         landRequestId: landRequest.id,
         pullRequestId: landRequest.pullRequestId,
-        landRequestStatus,
-        runningTargetingSameBranch,
+        // landRequestStatus,
+        // runningTargetingSameBranch,
         lockId,
       });
       return false;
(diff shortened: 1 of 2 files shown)


src/lib/Runner.ts
```diff
@@ -459,18 +456,51 @@ export class Runner {
...
         break;
     }
   };
 
+  // Mark the current request as failed / aborted and fail all dependents within 'status-transition' lock.
+  // So, multiple free slots are available in one go for the speculation engine to process.
+  failAllDependents = async (
+    currentLandRequestStatus: LandRequestStatus,
+    currentRequestState: IStatusState,
+    currentRequestReason: string,
```
Blasz, line 480: nit: The `current` prefix makes sense for the first parameter but I'm not sure about the other two since they represent the new state rather than the current state.

winniezheng (approved):

src/lib/Runner.ts
```diff
@@ -459,18 +456,51 @@ export class Runner {
...
+          landRequestId: currentLandRequestStatus.requestId,
+          pullRequestId: currentLandRequestStatus.request.pullRequestId,
+          dependentsPullRequestId: dependents.map((dependent) => dependent.request.pullRequestId),
+        });
+        for (const dependent of dependents) {
+          await this.failDueToDependency(dependent, [currentLandRequestStatus], lockId);
+        }
+      },
+      undefined,
+      100,
```
Blasz, line 500: I'm not sure that we should set a `ttl` here as the `next()` function itself doesn't have one and all the others that do have a `ttl` are separate interval functions that have a comment about not blocking `next`. I'd imagine we still want to be blocking `next` and so shouldn't set a 100ms `ttl`.

I don't quite understand the original reasoning for 100ms `ttl` anyway even after reading the comment, maybe @cheesehary recalls?

src/lib/__tests__/Runner.test.ts
Blasz, line 1: Can we add a test for this new behaviour?

src/lib/Runner.ts
GraceDmello: can rename them to `state` and `reason`

src/lib/Runner.ts
GraceDmello: Yes, it make sense for the interval functions to have lower `ttl`. In this case, we want to block the execution of the `next` functions.

Blasz (approved):

